PR #2, opened 2016-08-01
ASL becomes AML

PeterGV:
Clarify that Control Methods are written in ASL (not in AML), and only later *compiled* into AML.  A nit?  Perhaps.  But potentially confusing to a newb, and we ought to be clear when we can.


tedhudek:
Thanks for the PR, @PeterGV.  The update is live here:

https://msdn.microsoft.com/en-us/windows/hardware/drivers/acpi/evaluating-acpi-control-methods

PR #9, opened 2016-08-24
Update DIFx guidelines, document some INF syntax

jasonknichel:
clarify what it means to "use DIFxApp"


jasonknichel:
also added some edits to document some INF syntax

tedhudek:
Thanks @jasonknichel

PR #12, opened 2016-08-25
Update inf-manufacturer-section.md

jasonknichel:
fix the manufacturer section examples


tedhudek:
@jasonknichel: thanks again, this looks much better

PR #14, opened 2016-08-29
Update using-setupapi-to-uninstall-devices-and-driver-packages.md

jasonknichel:
fix formatting - add a space between two words


jasonknichel:
@tedhudek I think you overlooked this simple one the other day :)

tedhudek:
@jasonknichel thanks for the reminder.  I accepted but it looks like MSDN still renders without the space.  I'll rephrase to fix, likely tomorrow.  See https://msdn.microsoft.com/windows/hardware/drivers/install/using-setupapi-to-uninstall-devices-and-driver-packages.

PR #21, opened 2016-09-15
Reference ACPI 6.0 and add Type-C connector

MichelleBergeron:
Bug 7871805:"Using ACPI to Configure USB Ports on a Computer" to reference new version of ACPI specification


tedhudek:
@MichelleBergeron: looks great, thanks for the PR

Issue #23, opened 2016-09-22
Incorrect data in the Offloading the Segmentation of Large TCP Packets doc


I-gor-C:
Link: https://github.com/Microsoft/windows-driver-docs/blob/staging/windows-driver-docs-pr/network/offloading-the-segmentation-of-large-tcp-packets.md

Last paragraph states:

"Not set the **UserData** member of the NDIS_TCP_LARGE_SEND_OFFLOAD_NET_BUFFER_LIST_INFO structure when a send operation is complete. Instead, the miniport driver must set the LsoV2TransmitComplete.Reserved member to zero and the LsoV2TransmitComplete.Type member to NDIS_TCP_LARGE_SEND_OFFLOAD_V2_TYPE."

There is **_no_** UserData member in the NDIS_TCP_LARGE_SEND_OFFLOAD_NET_BUFFER_LIST_INFO structure.

PR #24, opened 2016-09-22
Remove bullet based on feedback

EliotSeattle:
Removed the last bullet point. We already say what the values should be set as depending on whether the request was v1 or v2. Thanks, @I-gor-C for the correction.

### DIFF
--- a/windows-driver-docs-pr/network/offloading-the-segmentation-of-large-tcp-packets.md
+++ b/windows-driver-docs-pr/network/offloading-the-segmentation-of-large-tcp-packets.md
@@ -105,7 +105,6 @@ In addition to the previous LSO requirements, LSOV2-capable miniport drivers mus
 
 -   Support TCP options, IP options, and IP extension headers.
 
--   Not set the **UserData** member of the [**NDIS\_TCP\_LARGE\_SEND\_OFFLOAD\_NET\_BUFFER\_LIST\_INFO**](https://msdn.microsoft.com/library/windows/hardware/ff567882) structure when a send operation is complete. Instead, the miniport driver must set the **LsoV2TransmitComplete.Reserved** member to zero and the **LsoV2TransmitComplete.Type** member to NDIS\_TCP\_LARGE\_SEND\_OFFLOAD\_V2\_TYPE.
 
  
 


I-gor-C:
Hi @EliotSeattle , now with the removal of this entire bullet there is no mentioning of what values to fill in LsoV2TransmitComplete in case of V2. With the current HLK version not filling anything causes the LSO test to fail (but writing 0 to the Reserved field as was written in the removed bullet resolves the failure). Can you please clarify?

tedhudek:
Hi @I-gor-C  , please see [NDIS_TCP_LARGE_SEND_OFFLOAD_NET_BUFFER_LIST_INFO](https://msdn.microsoft.com/library/windows/hardware/ff567882), which discusses what the values should be set as depending on whether the request is v1 or v2.  The protocol driver may not look at the Type field on completion, so meeting this requirement may not be necessary.  @EliotSeattle @pravb

I-gor-C:
Hi @tedhudek , thanks for the prompt reply. I have looked at the NDIS_TCP_LARGE_SEND_OFFLOAD_NET_BUFFER_LIST_INFO documentation - it explicitly mentions what to write in the LsoV1TransmitComplete for LSO v1 but for LSO v2 mode LsoV2TransmitComplete shouldn't be touched according to the documentation. However, not writing anything to the LsoV2TransmitComplete causes the LSO test to fail in HLK.

tedhudek:
@pravb can you help @I-gor-C ?

pravb:
The test should pass as long as the following are true:
LsoV2TransmitComplete.Reserved = 0
LsoV2TransmitComplete.Type = NDIS_TCP_LARGE_SEND_OFFLOAD_V2_TYPE

Can you check whether the driver is cloning or somehow causing these fields to have other values. Can you try explicitly setting these values before completion? An explicit set is what our virtual driver for hyper-v is doing so this should work.

I-gor-C:
Hi @pravb . The test indeed passes with that configuration. However, the bullet that was removed contained the only reference in the docs to this. Since the first part of the bullet was incorrect (the **UserData** field), I wasn't sure about the rest in that bullet - whether the test passed due to a side-effect or because it really requires the **Reserved** field to be 0 (it sounded weird that miniport driver should write into NDIS field marked as Reserved...). Now that you confirm that it actually needs to be 0, I suggest to return the removed bullet (without the part about **UserData**). And for consistency, the same info should be present in [NDIS_TCP_LARGE_SEND_OFFLOAD_NET_BUFFER_LIST_INFO](https://msdn.microsoft.com/library/windows/hardware/ff567882) as well.

Many thanks for your help!

tedhudek:
Thanks again @I-gor-C for caring about the docs.  It really makes a huge difference to have folks like you with an eagle eye.  And thanks @pravb for chiming in.

I've re-added the bullet to the bottom of the [source file](https://github.com/Microsoft/windows-driver-docs/blob/staging/windows-driver-docs-pr/network/offloading-the-segmentation-of-large-tcp-packets.md) with the modifications described above.  Please take a look and see what you think.  If you would like to make any changes, please feel free to submit a pull request right here.  Also see the [rendered page](https://msdn.microsoft.com/windows/hardware/drivers/network/offloading-the-segmentation-of-large-tcp-packets).

Fixes #23

tedhudek:
BTW, [NDIS_TCP_LARGE_SEND_OFFLOAD_NET_BUFFER_LIST_INFO](https://msdn.microsoft.com/library/windows/hardware/ff567882) is in our older, slower authoring environment.  The update should appear there midday Thursday PT.

I-gor-C:
Looks good, thanks a lot!

tedhudek:
You can now see the update to https://msdn.microsoft.com/library/windows/hardware/ff567882, specifically the description of the **LsoV2TransmitComplete** member.

PR #29, opened 2016-10-27
Adding link to accessing keys from WDF drivers

Shyamal-V:
Adding super helpful link to accessing keys from WDF drivers


tedhudek:
Thanks for the link, Shyamal.  This should be live in a couple minutes.

PR #30, opened 2016-10-27
Update exploring-the-windows-vista-audio-engine.md

danteliujie:
A simple fix about the word should be 
"in shared mode or exlusive mode" 
instead of
"in shared mode ore exlusive mode"


tedhudek:
Thanks, Dante, for finding and fixing this!  We really appreciate it.

PR #43, opened 2016-12-13
Update installing-and-upgrading-network-components.md

jtippet:
fix link-  I think the page intended to point to the high-level overview, not the detail page of the same name


tedhudek:
Verified, this was a mistake in the original (pre-OP) source.  Thanks for spotting it!

PR #44, opened 2016-12-14
Rephrase the virtualization disabling description

I-gor-C:


tedhudek:
Thanks for the PR, Igor!

PR #47, opened 2017-01-06
Fix typo in changing-the-friendly-name-of-a-boot-entry.md

MattKotsenas:
Fix a typo where Windows 10 was referred to as Windows XP in changing-the-friendly-name-of-a-boot-entry.md


tedhudek:
Thanks for the PR, Matt!

PR #50, opened 2017-02-08
fix formatting of FSFilter class list

jasonknichel:


tedhudek:
Thanks @jasonknichel the updates are live at https://msdn.microsoft.com/en-us/windows/hardware/drivers/ifs/file-system-filter-driver-classes-and-class-guids

PR #52, opened 2017-02-09
Split IOCTL links onto separate lines.

cburg:
Links to the IOCTL documentation were all showing on one line and looked like a single large macro instead of three individual macros.


tedhudek:
Thanks @cburg looks much better now!  https://msdn.microsoft.com/en-us/windows/hardware/drivers/storage/supporting-mount-manager-requests-in-a-storage-class-driver

PR #54, opened 2017-02-10
Update the topic for filtering memory mapped io

aktsuda:
- Adding the general way to filter memory mapped I/O as filtering paging I/O.
- Replace "the IFS documentation in the WDK" with "NTFSD (Windows File System Developers List) newsgroup on OSR" because no one can find such information in WDK.
I discussed this with Christian Allred. So, he or his team might review this.


tedhudek:
Thank you for the PR, Akihisa-san @aktsuda

PR #60, opened 2017-02-18
Update offloading-the-segmentation-of-large-tcp-packets.md to cover a case when LSO is turned off

Ray-Cai:
Added documentation for a corner case where if LSO is turned off on the miniport, the miniport shall drop all packets with any non-zero LSOv1 or LSOv2 OOB data.


tedhudek:
Thanks for the PR @Ray-Cai the updates are now live at https://msdn.microsoft.com/en-us/windows/hardware/drivers/network/offloading-the-segmentation-of-large-tcp-packets

PR #61, opened 2017-02-23
Update the-sample-driver-setup-information---inf--file.md

rajban2017:
VSO Bug#: 11035923


tedhudek:
Thanks for the PR @rajban2017 , this change is now live at https://msdn.microsoft.com/en-us/windows/hardware/drivers/portable/the-sample-driver-setup-information---inf--file

PR #73, opened 2017-03-31
fix some small coherency problems

jasonknichel:


tedhudek:
Thanks for the sharp eye as always @jasonknichel this should be live within 10 minutes

PR #74, opened 2017-04-04
Update sensor-batching-for-power-saving-.md

JocelynBerrendonner:
Fixed the name of the PKEY_Sensor_WakeCapable property (no underscore between Wake and Capable) as exposed in the WDK/SDK.


tedhudek:
Thanks @JocelynBerrendonner great find.

PR #81, opened 2017-04-13
link to what is meant by destination directory

jasonknichel:


tedhudek:
Thanks @jasonknichel this should be live within a few minutes

PR #94, opened 2017-04-25
Fix markdown format.

KexyBiscuit:
This commit fixes `Universal Serial Bus (USB)`'s link.


tedhudek:
Thanks @KexyBiscuit for the sharp eye!

PR #96, opened 2017-05-03
Corrected bad link to inf2cat

iliketoprogram14:


tedhudek:
Thanks @iliketoprogram14 this should be live in a couple minutes

PR #102, opened 2017-05-09
Fix formatting of word

hpsin:


tedhudek:
Thanks @hpsin for your sharp eye!  The update should appear on the live site momentarily.

PR #109, opened 2017-05-16
No need to escape path separator in INF

AndrewGaspar:


tedhudek:
Thanks @AndrewGaspar this should be on the live site in a few minutes

PR #126, opened 2017-05-25
Update windows-threshold--what-s-new-for-audio.md

MartinGPuryear:
Two small typos ('ion' => 'in'; 'users can the' => 'users can then').


tedhudek:
@MartinGPuryear thanks for the PR's.  Changes should be live within a few minutes

PR #129, opened 2017-05-29
Update example-15--measuring-dpc-isr-time.md

I-gor-C:
set correct filename extension for the html output file


tedhudek:
Thanks @I-gor-C another good find!

PR #132, opened 2017-05-30
Add a note about an edge case for query remove failed.

jasonknichel:


tedhudek:
Thanks @jasonknichel this will be live in a couple minutes

PR #146, opened 2017-06-09
Remove heading redundancy

sethgr:
Describing a 'best practice' as optional seems redundant.


tedhudek:
Thanks @sethgr always good to hear from you.  Appreciate your taking a look at this

PR #155, opened 2017-06-16
IpChecksumFailed should be set if checksums fail validation

pravb:
IpChecksumFailed should be set if checksums fail validation, the text says incorrectly that IpChecksumSucceeded should be set.


barrygolden:
Thanks for the update **@pravb**, the change should be live soon.

PR #160, opened 2017-06-19
Update makecert-test-certificate.md

veleek:
Adding a space between CryptoAPI and MakeCert in the first paragraph.


tedhudek:
Thanks @veleek good find

PR #162, opened 2017-06-21
Fixed minor typo

goeiecool9999:
Significant advantageous was written instead of advantages.


tedhudek:
Thanks for the sharp eye @goeiecool9999 this change will be live within a few minutes

PR #166, opened 2017-06-23
Update early-launch-antimalware.md

nbassett-MSFT:
Added reference to Windows 10
Adjusted anti-malware to antimalware, to match existing documentation we provide to partners
Tried to put the other ELAM links on separate lines, not sure if I did that properly.


tedhudek:
Thanks for the PR @nbassett-MSFT this should be live within a few minutes

PR #167, opened 2017-06-23
Update elam-prerequisites.md

nbassett-MSFT:
Clarified that the HCK tests only apply to pre-win10 submissions
Updated the submission section to refer to the new submission page, and highlighted the mvi/via requirement for submission


barrygolden:
Thanks for the PR @nbassett-MSFT this should be live now

PR #168, opened 2017-06-26
Add more details about rss processor number & numa number

jtippet:


barrygolden:
Thanks for the PR @jtippet your changes are now live

PR #170, opened 2017-06-28
Update elam-driver-requirements.md

nbassett-MSFT:
I added a link at the top to the sample code we provide on github for an ELAM driver.  I'm not sure if this should also be on the main summary page for the feature as well (open to suggestions).  It's important to call out because I think it will prevent a lot of development questions.


barrygolden:
Thanks for the PR @nbassett-MSFT your change is now live

PR #171, opened 2017-06-28
Update using-an-extension-inf-file.md

NabilFates:
Added example to clarify the ordering case.


barrygolden:
Thanks for the PR @NabilFates your changes are now live

PR #172, opened 2017-06-28
Update universal-driver-scenarios.md

NabilFates:
Adding a bit of text about ordering. The information might be duplicated from the main extension INF article, but we feel that the clarification/reminder will be useful to some developers.


barrygolden:
Thanks for the PR @NabilFates your changes are now live

PR #177, opened 2017-07-06
Update specifying-virtual-wifi-bus-dependencies.md

zlockard:
Small changes to be consistent with the typical way we refer to INF sections


duncanmacmichael:
Thanks for the clarification on terminology, @zlockard. Your PR will go live soon.

PR #181, opened 2017-07-11
Updates to partner settings apps

allisonhopper:
Updating this page with the new content from the UX WEG.


aahill:
Thanks Allison, I'll take a look at this.

aahill:
I've cloned this to a local branch and will take it from here. Thanks again!

PR #184, opened 2017-07-13
fix the link to device manager errors

jasonknichel:
Don't hardcode en-us in the link


tedhudek:
Thanks for the PR @jasonknichel your change will be live momentarily

PR #185, opened 2017-07-20
Fix symbol-path.md style formatting markdown issue

kiazhi:
Fixed inline issue #186 due to Markdown language syntax issue.
Instead of using \* asterisk symbol to highlight syntax, I have replaced it using \` grave accent to highlight code syntax.

### DIFF
--- a/windows-driver-docs-pr/debugger/symbol-path.md
+++ b/windows-driver-docs-pr/debugger/symbol-path.md
@@ -26,7 +26,7 @@ The debugger's symbol path is a string that consists of multiple directory paths
 
 Relative paths are supported. However, unless you always start the debugger from the same directory, you should add a drive letter or a network share before each path. Network shares are also supported.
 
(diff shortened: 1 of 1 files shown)


DOMARS:
I think this is an improvement, thanks.

tedhudek:
Thanks @kiazhi this change should be live soon

kiazhi:
Hi @tedhudek, 

Checked the live docs now. Looks a lot better and easier to read now.

Since this is not part of PowerShell-Docs repo, can I check with you if it is normal to have markdown headings containing html `<span></span>` tag?

```html
## <span id="Symbol_Path_Syntax"></span><span id="symbol_path_syntax"></span><span id="SYMBOL_PATH_SYNTAX"></span>Symbol Path Syntax
```

Does DOCS automatically generate the section links on "In this article" right panel with normal markdown headings or does it requires the `<span></span>` tag?

tedhudek:
Hi @kiazhi, I think the span tags are largely an artifact of the old XML based docs we migrated out of.  They might be used for linking to subsections of pages, but Docs does auto generates the sidebar sections links from the markdown titles.  In short it's probably not a problem to remove them.  Thanks again for caring about the docs!

PR #187, opened 2017-07-20
Update elam-prerequisites.md

nbassett-MSFT:
Adding URLs for VIA, MVI, and making the mvi@microsoft.com address clickable.


tedhudek:
Thanks Nick this change will be live soon

PR #188, opened 2017-07-20
Formatting fixes for virtual-hid-framework--vhf-.md

mikemaksymowych:
Fix code section in "Submit the HID input report" to include the function header
Remove unnecessary blank lines
Remove unnecessary spaces at the end of lines


tedhudek:
Thanks @mikemaksymowych this change will be live momentarily.  I also fixed an error on our end, namely that the fenced code blocks were indented, which at least in VS Code causes the triple-tick to appear as part of the rendered code block.  The fix was just to unindent the code blocks.  Thanks for the sharp eye

PR #193, opened 2017-07-27
Remove DisableProtectedAudioDG section

nehinnan:
On RS1+, AudioDG runs unprotected. This regkey is no longer required to test unsigned APOs.


tedhudek:
Thanks for the PR @nehinnan this should be live within minutes

PR #194, opened 2017-07-28
Remove bad guidance for hiding devices in device manager.

jasonknichel:
That guidance is partly out of date and partly wrong.  A device should be in a device setup class that makes sense for that device.  It shouldn’t be put in some random class just because that class is marked as NoDisplayClass.  Also, we highly discourage INFs defining new classes and that isn’t allowed in universal INFs.  So the documentation shouldn’t recommend that either.


tedhudek:
Thanks for the PR @jasonknichel this should be live within minutes

PR #198, opened 2017-08-03
Update setting-up-a-connection-to-a-virtual-machine-in-visual-studio.md

kevinchannon:
Correcting instruction 1 in the  "Starting the debugger" section to say that the "Attach to process" option is in the "Debug" menu, instead of the "Tools" menu


tedhudek:
Thanks @kevinchannon for the PR, this change should be live within a few minutes

PR #200, opened 2017-08-07
Fix small typo

modi1123:


DOMARS:
Looks like a typo that should be fixed, thanks!

PR #206, opened 2017-08-17
Update bug-check-0x6--invalid-process-detach-attempt.md

jasonfish4:


tedhudek:
Thanks @jasonfish4 for the PR, your change will be reflected on the live page momentarily.

PR #209, opened 2017-08-18
Remove Important note about Sysdev

desowin:
According to https://sysdev.microsoft.com/en-US/Hardware/CompatibilityPlaylists/DevCenter.aspx this important note seems to be obsolete as of 01/17/2017. This has caused some unneccessary confusion in my case as due to this note I though I have to create Sysdev account. Creating new accounts using company email addresses in Sysdev is complicated due to cleaning up the Azure/Microsoft account overlap. I believe removing this obsolete warning will save others from this confusion.


aahill:
Thanks for catching this; sorry about the confusion! This feature has indeed been ported over from Sysdev, so I'll merge the change in!

PR #210, opened 2017-08-19
Update accessing-user-space-memory.md

jasonfish4:
Updated information that explains how drivers can access and allocate memory in user-mode without requiring a thread to be attached in the user-mode address space.


PageWriter-MSFT:
Hi @jasonfish4 

Thanks for your contribution. 

Unfortunately, we are unable to accept the update. The use of MmCopyVirtualMemory is not supported. Because of your contribution, we have updated the first paragraph to clarify this topic.

Thanks,

WDK content writer

PR #214, opened 2017-08-29
clarify some extension INF details

jasonknichel:


tedhudek:
Thanks @jasonknichel this is live.  I also added a graphic from Nabil to clarify that [section](https://docs.microsoft.com/windows-hardware/drivers/install/using-an-extension-inf-file#specifying-extensionid).

PR #219, opened 2017-08-30
clarify INF encoding requirement

jasonknichel:


tedhudek:
Thanks @jasonknichel this is now live at https://docs.microsoft.com/en-us/windows-hardware/drivers/display/general-unicode-requirement

PR #220, opened 2017-09-05
Update bug-check-0x7e--system-thread-exception-not-handled.md

wm1:
The extra "(Developer Content)" breaks index in chm file. Other bug check codes does not have this word.


tedhudek:
Thanks @wm1.  I wish we could rebuild the .chm for you but it's not so easy to make that format now.  To get the update, you could try downloading the pdf from https://docs.microsoft.com/windows-hardware/drivers/debugger/bug-check-0x7e--system-thread-exception-not-handled (see PDF button in lower left)

wm1:
WinDbg still ships chm as its help format, and that is where I noticed the index problem.

PR #222, opened 2017-09-07

lukangel:
Updates to the SoftwareType 1 section per Philippe's email


tedhudek:
Thanks @lukangel this will be live momentarily

PR #231, opened 2017-09-15
update HKR table

jasonknichel:


tedhudek:
Thanks again @jasonknichel these should both be live momentarily

PR #234, opened 2017-09-18
fix spacing between command and options

goldstar611:
Whitespace adjustments only, No tests conducted


tedhudek:
Thanks for the PR @goldstar611 your update should be live momentarily

PR #238, opened 2017-09-26
Fix '\<' formatting

nitoyon:
Fix formatting with `\<`.

(ex) https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/gflags-details

image


tedhudek:
Thanks @nitoyon I verified that all of your fixes render properly on the site.  Thanks again for taking the time to do this.

PR #246, opened 2017-10-02
Update driver store modification restrictions

jasonknichel:


tedhudek:
Thanks @jasonknichel, this will be live momentarily

PR #247, opened 2017-10-02
Fix typo

I-gor-C:


tedhudek:
Thanks @I-gor-C this will be live in a moment

PR #249, opened 2017-10-05
update interface sample

jasonknichel:
A new interface can be added to the system between the call to get the size of the list and the call to get the list.  Code should be resilient to that happening.


tedhudek:
Thanks @jasonknichel this should be live momentarily.

PR #250, opened 2017-10-05
Update native-objects-in-javascript-extensions.md to fix formatting

nitoyon:
Fix https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/native-objects-in-javascript-extensions


tedhudek:
Thanks @nitoyon as always

PR #251, opened 2017-10-09
Update Windows10SDriverRequirements.md

Cymoki:
was requested to make the changes above to align with what fall creator update will be enforcing.


tedhudek:
Thanks for the PR @Cymoki this should be live within a few minutes.

PR #256, opened 2017-10-18
Wrong directive stated in text in inf-addsoftware-directive.md

eddiemon:
Change incorrect "CopyINF" to correct "CopyFiles" directive in text. Link is OK


tedhudek:
Thanks @eddiemon for your sharp eye!  This change will be on the live page within a few minutes.

PR #257, opened 2017-10-20
Update automatic checks introduction

dkluemke:
Update the automatic checks introduction to indicate that most automatic checks have been refactored into the DV standard flags.


tedhudek:
Thanks @dkluemke for the PR, your update will be live within a few minutes

PR #262, opened 2017-10-26
Update manage-your-hardware-submissions.md

NabilFates:
Adding register ID SLA.


tedhudek:
Thanks @NabilFates this will be live momentarily.

PR #263, opened 2017-10-30
Update audio-processing-object-architecture.md

kkotte:
Added clarification to the note about streams that share the same MFX. The existing documentation led to a misinterpretation that there could only be a single instance of the MFX globally.


DOMARS:
Looks good to me, thx for checking.

PR #264, opened 2017-11-01
Update printer-inf-file-entries.md

Tauders:
Added markdown for the table of the INF file entries.


barrygolden:
Thanks for the PR, @Tauders. Your changes will be live soon.

PR #265, opened 2017-11-01
Update using-an-extension-inf-file.md

NabilFates:
Some partners are interpreting this as "we need to register an extension ID for every OEM variant of the extension INF". Adjusting the wording a little to help clarify.


tedhudek:
Thanks @NabilFates

PR #266, opened 2017-11-01
Update manage-your-hardware-submissions.md

NabilFates:
Not every extension INF submission will require to register an extension ID. Removing the word "submission" to try and make this clear.


aahill:
Thanks Nabil, 

the change looks good - should be live soon!

PR #278, opened 2017-11-14
Added Comodo Code Signing Links

computeronix:
Added Comodo for Code Signing and EV Code Signing (had a customer ask about their support)
Note Please change the comodo links to go.microsoft.com links as I do not have access to add go.microsoft.com links


aahill:
Thanks @computeronix, I've added the links, and will merge this into master. Should be visible in a day!

PR #286, opened 2017-11-23
Update create-a-new-hardware-submission.md

aktsuda:
Adding notes to "Create a new hardware submission" to avoid the issue that Dev Center includes x86 signatures in x64 signatures due to a HLK bug. (See OS Bug 14567890.)


tedhudek:
Thanks @aktsuda and @Cymoki these changes are live now.

Issue #45, opened 2016-12-19
Contradiction in network interface provider documentation


I-gor-C:
[Registering as interface provider](https://msdn.microsoft.com/windows/hardware/drivers/network/registering-as-an-interface-provider) explicitly states that any network driver type can serve as an interface provider:
> ... _However, miniport drivers and filter drivers can also be interface providers_

On the other hand, [OID_GEN_ADMIN_STATUS](https://msdn.microsoft.com/en-us/library/windows/hardware/ff569437(v=vs.85).aspx) documentation states the opposite: 
> _Only NDIS network interface providers, and therefore **not miniport drivers or filter drivers**, must support this OID as an OID request_. 

Actually, seems like the latter appears in the descriptions of all the [NDIS Network Interface OIDs](https://msdn.microsoft.com/en-us/library/windows/hardware/ff566545(v=vs.85).aspx)

tedhudek:
Investigating...

jtippet:
The first quote means:  A driver (.sys file) that is an NDIS miniport driver can also register as an interface provider.  In that case, the .sys file is both a miniport and an interface provider.

The second quote means that `OID_GEN_ADMIN_STATUS` (and similar OIDs) are not part of the contract for miniport drivers.  Your `MiniportOidRequest` handler will never receive `OID_GEN_ADMIN_STATUS`.  However, if your .sys file hosts both a miniport driver and an interface provider, then the interface provider aspect of your code needs to implement `OID_GEN_ADMIN_STATUS`.

It's worth pointing out that interface providers are niche stuff.  I'm not aware of any 3rd-party drivers that register as interface providers.  You can build one if you want, but 99% of cases are handled by the built-in interface providers.  Interface providers (and the OIDs that belong exclusively to them, like OID_GEN_ADMIN_STATUS) are typically not interesting and won't solve most problems.

I-gor-C:
Thank you Jeffrey. I think it's worth adding this clarification to the documentation to avoid confusion.

One thing still isn't clear to me, though - what is the meaning on "interface" in this context? Is registering an interface is something like installing a new adapter which appears in the network connections (such as the loopback adapter)? I haven't found any guide about interaction with the registered interface.

jtippet:
An "interface" is meant to correspond to [RFC 2863](https://tools.ietf.org/html/rfc2863).

NDIS has a built-in interface provider that automatically provides interfaces for:

- Every miniport driver
- Every filter (LWF) module
[...]

PR #287, opened 2017-11-24
Update oid-gen-media-connect-status.md

I-gor-C:
Indications description addressing only NDIS 5.1


I-gor-C:
Hi Duncan,
I was looking for the original OID_GEN_MEDIA_CONNECT_STATUS. The OID_GEN_MEDIA_CONNECT_STATUS_EX seems to be related only to connection status of interface providers rather than to the common case. 

P.S. About a year ago @jtippet has clarified the usage of network interfaces in https://github.com/MicrosoftDocs/windows-driver-docs/issues/45

duncanmacmichael:
Okay, that makes sense. I'll go ahead and merge this then. Thanks!

PR #288, opened 2017-12-01
Update get-drivers-signed-by-microsoft-for-multiple-windows-versions.md

soul4soul:
URL was pointing to an MSDN page which didn't redirect to the correct doc's url.


tedhudek:
Thanks @soul4soul  good eye!

PR #290, opened 2017-12-06
Update create-a-system-settings-application.md

brentkendall:
Added missing word "be" to "all links must be relevant to the page"


tedhudek:
Thanks @brentkendall for the sharp eye.  This change will be live momentarily

PR #291, opened 2017-12-07
Using the kernel stack

wm1:


tedhudek:
Thanks for the PR @wm1 this will be live in a few minutes

PR #293, opened 2017-12-11
Update -fileobj.md

alfredmyers:
Are there any plans to move documentation such as for the FILE_OBJECT structure from MSDN to docs.microsoft.com?
I didn't find anything on docs, so I created a link pointing to MSDN.


tedhudek:
Thanks @alfredmyers - we're still actively moving content from MSDN (and TechNet) to Docs, but I don't have a date I can share as to when the FILE_OBJECT page might be moved.

PR #295, opened 2017-12-14
Update inf-reboot-directive.md

NabilFates:
This directive can be used in a Universal INF. Fixing the note.


tedhudek:
Thanks @NabilFates

PR #297, opened 2017-12-19
Minor fix

alfredmyers:


EliotSeattle:
Thanks for taking the time to do this, @alfredmyers.

PR #298, opened 2017-12-27
Naming tweaks

alfredmyers:


tedhudek:
Thanks @alfredmyers for fixing these.  Fixes should be live momentarily

PR #299, opened 2017-12-28
Make the list of links a real markdown list

timrprobocom:


tedhudek:
Thanks for the PR @timrprobocom your update will be live momentarily

PR #300, opened 2017-12-28
Fix for rule difference

timrprobocom:
The microsoft.com markdown rules are slightly different than the GitHub markdown rules.  My last commit had a glitch on the live page.  Sorry.


tedhudek:
Thanks @timrprobocom yes, the rendering engines are different, no way you could know or test :)

timrprobocom:
Still, this process works amazingly well, and quickly, too.  My compliments.

tedhudek:
Thanks, we like it too.  We're still actively migrating content into the Markdown/Git-based system, so let us know if you come across a page you want to update that doesn't have an edit button.  Quick way to tell is if the page is on MSDN, it's the old system, and if it's on docs.ms.com, then it's on the new

PR #302, opened 2017-12-30
Update windows-compatible-hardware-development-boards.md

chsha:
Updated link


tedhudek:
Thanks for the PR @chsha this change will be live momentarily

PR #303, opened 2018-01-08
Fix typo

0xbadfca11:


tedhudek:
Thanks @0xbadfca11 this change should be live momentarily

PR #306, opened 2018-01-10
Update implementing hw offloaded apo effects

nehinnan:
Address language feedback re: offload pin instance stream mixing from bug 14287522.


DOMARS:
These changes look good to me. Thanks for making the update.

tedhudek:
Thanks @nehinnan this change should be live momentarily

PR #307, opened 2018-01-11
Providing clarity on the threat model and intersection with MOR's ACPI _DSM method

out0xb2:
See subject line


barrygolden:
Thanks @out0xb2, This change should be live soon.

PR #308, opened 2018-01-12
Adds link to the Debugging Tools for Windows page

AmadeusW:
The linked page contains download links. I'm not sure if I should have used an absolute URL or relative URL (just up the directory: `..`)


tedhudek:
Thanks for the PR @AmadeusW -- you can use either absolute URL or relative directory path.  I remove the 'en-us' from the URL so user's browser settings can make that choice.  Otherwise looks great!  Will be live momentarily.

DOMARS:
Looks good thanks for the contribution!

PR #311, opened 2018-01-17
Update index.md

hraom:
Please hyperlink Bug Management - Migrated to http://aka.ms/collaboratedocs


tedhudek:
@aahill you might consider putting HDC in parentheses in the sentence preceding the table, so it's clear to readers that it corresponds to Hardware Dev Center

aahill:
@tedhudek thanks for the tip; went ahead and did that. @hraom Thanks for the pull request! This information should be live shortly.

PR #312, opened 2018-01-17
typo: it's -> its

jtippet:


aahill:
@jtippet good catch! Thanks for the pull request, the change should be visible soon.

PR #313, opened 2018-01-18
Change cross-certificate list format to fix broken display

lepaperwan:
This proposed change moves the list of cross-certificates to a table in order to ensure consistent alignment of CA name, certificate thumbprint and download link as opposed to its current state which turns thumbprint, download link and the name of the next CA in a oneliner.

Download links have been stripped of the CA name and "2 KB zip file" mention to shorten lines and avoid line breaks within cells on smaller screens.


tedhudek:
Thanks @lepaperwan for making this change.  The result is much clearer!  The update is live at https://docs.microsoft.com/en-us/windows-hardware/drivers/install/cross-certificates-for-kernel-mode-code-signing

PR #315, opened 2018-01-23
Update desktop-cosa-apn-database-settings.md

wjendli:
Add esim provisioning to purpose groups


duncanmacmichael:
Hi @wjendli , thanks for the PR! Everything looks good and should be live soon.

PR #317, opened 2018-01-26
More updates to language regarding stream mixing in the audio driver

nehinnan:


DOMARS:
@tedhudek looks good to me, thx.

tedhudek:
Thanks @nehinnan your change will be live by eod

PR #320, opened 2018-01-29
Update oid-gen-maximum-frame-size.md

I-gor-C:
Remove typo


duncanmacmichael:
Thanks for the catch, Igor. I'll merge the PR for you and it should be live soon!

PR #328, opened 2018-02-06

NabilFates:
Adding clarity to the note on uninstallation.


tedhudek:
Thanks @NabilFates this is live at https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-addsoftware-directive#remarks

PR #329, opened 2018-02-08
Update hardware-certification-submissions.md

SillyKeith:
updated symbols wording to say we want public not private symbols and a link to how to create public symbols


tedhudek:
Thanks @SillyKeith, will be live momentarily.

@aahill check it out for awareness, I went ahead and merged it in.

PR #330, opened 2018-02-10
Update evt-net-adapter-preview-wake-pattern.md Clarify further that PatternToBeAdded should not be referenced beyond the callback.

Shyamal-V:
Clarify further that PatternToBeAdded should not be referenced beyond the callback.


duncanmacmichael:
Hi @Shyamal-V , good catch. That clarification was there before, but it was down in the remarks section. I think you're right in that it should be moved up to the parameter description for further emphasis, so I went ahead and removed the paragraph from the remarks section in addition to your change. I'll go ahead and merge the PR. Thanks!

duncanmacmichael:
Ah, I gotcha. Sure, I can add it back in so we have it in both places. Thanks!

PR #333, opened 2018-02-14
Include image file format.

OswaldoSantana:
This is to avoid consumers wondering which file format should I use.


duncanmacmichael:
Hi @OswaldoSantana , thanks for the clarification on the file type. I'll go ahead and merge this PR and it should be live shortly.

PR #334, opened 2018-02-14
Remove UMDF V1 references

Shyamal-V:
UMDF V1 is not supported for one core based systems and we're discouraging its use in favor of UMDF v2. Remove references to UMDF V1 from GPIO documentation


tedhudek:
Thanks @Shyamal-V sorry for the delay, should be live momentarily

PR #337, opened 2018-02-15
make the echo-driver step-by-step doc more precise

noguxun:
there are some format issues (no line break). also added steps missed for windows 10.


DOMARS:
@noguxun thanks for this input. It may take me a bit to double check, but I appreciate your eye for adding additional detail to this lab.

DOMARS:
@noguxun I added text to address the issues that you raised that should make the lab even easier to follow. Those changes are now live:

https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/debug-universal-drivers---step-by-step-lab--echo-kernel-mode-

Thanks again for this input.

PR #339, opened 2018-02-21
Fix broken app verifier links

Shyamal-V:


tedhudek:
Thanks @Shyamal-V sorry for the delay, should be live momentarily

PR #340, opened 2018-02-26
fix formatting of list of classes

jasonknichel:


tedhudek:
Thanks @jasonknichel this should be live in minutes

PR #343, opened 2018-02-28
fix link: API validation

charliearcuri:
should point to page on ApiValidator.exe not a page on InfVerif.exe


tedhudek:
Thanks @charliearcuri good find. This will be live soon

PR #363, opened 2018-03-14
Need to update .dll version in discription 

mikeman06:
If you follow the "Windows API sets" link then chose Windows 8.1 (https://msdn.microsoft.com/en-us/library/windows/desktop/dn933214) you won't find api-ms-win-devices-config-l1-1-0.dll Looks like it's been updated to 1-1-1 which is what I want to update.


tedhudek:
Thanks @mikeman06, the update should be live momentarily

PR #371, opened 2018-03-19
Translates into korean.

somma:
This PR translates two docs into Korean.

- kernel\previousmode.md
- kernel\using-nt-and-zw-versions-of-the-native-system-services-routines.md


tedhudek:
Hi, thanks for the PR.  Sorry, but we can only take contributions in English.  You can see a translated page if you change "en-us" in the URL to "ko-kr", but the "ko-kr" page does not have an Edit button currently.  Apologies, and thanks again for taking the time to submit this PR.

PR #394, opened 2018-03-26
Grammer correction

NazmusLabs:
There was a grammatical error in the first sentence. Change "use" to "used"

### DIFF
--- a/windows-driver-docs-pr/mobilebroadband/cosa-overview.md
+++ b/windows-driver-docs-pr/mobilebroadband/cosa-overview.md
@@ -11,7 +11,7 @@ ms.technology: windows-devices
 
 # COSA overview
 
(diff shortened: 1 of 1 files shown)


duncanmacmichael:
Hi @NazmusLabs , thanks for the feedback. We appreciate the sharp eye. With respect to that verb, "use" is correct in this case as the topic is meant to describe a present-tense, ongoing process. To simplify things, you could read the sentence as "This is what MOs use in Version 1703 and later to provision Windows."

Does that help?

duncanmacmichael:
Hi @NazmusLabs , I'll go ahead and close the PR now. Thanks again for the contribution!

NazmusLabs:
Greetings, @duncanmacmichael! Thanks for your reply, but the word "used" is correct here because it is not being applied in past tense form; rather, it is being is saying that "MOs are being USED by another component". Note the "are being", which is in present tense. Used here is a possessive form, and not past tense. It should be used.

PR #406, opened 2018-04-02
Update where-setup-searches-for-drivers.md

SillyKeith:
Intel had questions about Critical vs Optional updates and it makes sense to "Document" how this works.


tedhudek:
Thanks for the PR @SillyKeith

PR #407, opened 2018-04-02
Update extension-inf-files.md

NabilFates:
Removing duplicate question.


aahill:
Thanks Nabil, I've merged this into master. Should be live soon!

PR #413, opened 2018-04-03
Update tracelog-command-syntax.md

RongshengXu:
Update the usage of tracelog to stop a global logger session.


duncanmacmichael:
Hi @RongshengXu , thanks for the clarification. I'll go ahead and merge your PR so it should be live soon.

PR #419, opened 2018-04-07
Grammer: possessive form

NazmusLabs:
Updated the sentence: "COSA[…]is the new format Mobile Operators (MOs) 'use' in Windows 10, version 1703 and later to provision Windows devices for mobile broadband." Changed the "use" to "used".

This is not a past tense. It is a possessive form. Something is USING Windows 10 to accomplish a task, in this case to provision devices for mobile broadband.

Let's give a couple examples. "This car is used by Emily for her daily commutes". This sentence is still in present tense. To say "This car is use by Emily for her daily commutes" would be grammeritacally incorrect.

Another example: "Visual Studio 2017 is used for developing UWP apps". This is a present tense, but, again, note the "used" is correct here. If it were past tense, it would be something like this: "Visual Studio 6.0 was used for C++ and VB6 development in the 90s". Note that the ver "to be" is either "is" or "was", depending on whether the sentence is past or present and the verb "to use" is alwasy written as "used" in both cases.

### DIFF
--- a/windows-driver-docs-pr/mobilebroadband/cosa-overview.md
+++ b/windows-driver-docs-pr/mobilebroadband/cosa-overview.md
@@ -11,7 +11,7 @@ ms.technology: windows-devices
 
 # COSA overview
 
-COSA, or Country and Operator Settings Asset, is the new format Mobile Operators (MOs) used in Windows 10, version 1703 and later to provision Windows devices for mobile broadband. The existing APN database (apndatabase.xml) from Windows 8, Windows 8.1, and versions of Windows 10 before Windows 10, version 1703 has been converted to COSA, which is ingestible by the new provisioning framework. Note that these previous versions of Windows will continue to use the older APN database for provisioning desktop devices.
+COSA, or Country and Operator Settings Asset, is the new format that Mobile Operators (MOs) use in Windows 10, version 1703 and later to provision Windows devices for mobile broadband. The existing APN database (apndatabase.xml) from Windows 8, Windows 8.1, and versions of Windows 10 before 1703 has been converted to COSA, which is ingestible by the new provisioning framework. Note that these previous versions of Windows will continue to use the older APN database for provisioning desktop devices.
 
 For more information about the older APN database, see [APN database overview](apn-database-overview.md).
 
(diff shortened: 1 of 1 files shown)


duncanmacmichael:
I see what you're saying, @NazmusLabs and I appreciate your diligence. :-) I think my original sentence would indeed be clearer if I added a qualifier word such as "that" or "what". For instance, "COSA is what MOs use in Version 1703 and later..." or "COSA is the new format that MOs use in Version 1703 and later...". At least, that was my intention. Since I didn't use the word "by," the verb "use" was transitive; that is, the sentence's voice was active rather than passive because the subject performing the action immediately preceded the verb. If it were passive, like your second example ("is used by"), then "use" would indeed be incorrect. I probably wrote it that way and omitted the helper word "that" unconsciously because I prefer writing with the active voice if possible.

Anyway, your point is absolutely correct in that it should be clarified and consistent with our other documentation. Thanks for your passion in improving the docs. Which of these do you think is better?

1.  COSA, or Country and Operator Settings Asset, is the new format used by Mobile Operators (MOs) in Windows 10, version 1703 and later to provision Windows devices for mobile broadband.
2. COSA, or Country and Operator Settings Asset, is the new format that Mobile Operators (MOs) use in Windows 10, version 1703 and later to provision Windows devices for mobile broadband.

NazmusLabs:
Hi, and thanks again for the response. Now that I understand what was intended on the original sentence, and you are absolutely correct. The use of "that" really makes it unambiguous. And, I will admit, it is still correct without the "that", so the sentence had no grammatical errors; so, my apologies. :) It seems that the lack of "that" made it ambigious for me.

As for which one is better, I'd go with the latter (number 2) because it requires minimal changes to the sentence and uses less words to describe the same thing. Also, active form just sounds better here than the passive form, which is something a certain Microsoft always used to nag me about: "Passive form, consider revising." ~Signed, Clippy from MS Word. :)

duncanmacmichael:
My pleasure, @NazmusLabs . No need to apologize; I'm happy we could have a good grammatical conversation and improve the docs at the same time. :-) Thanks for helping disambiguate the sentence! It'll help me be aware of such things in the future, too. I'll make the change and push it live.

PR #424, opened 2018-04-11
Update create-a-new-hardware-submission.md

SillyKeith:
Removing the Is Universal Checkbox option as it's gone.    We also need a new screen shot that does not show the "Is Universal Checkbox"  (images/drivers-certification.png)


aahill:
Thanks Keith! I was able to modify the existing image to remove the checkbox. These changes should be visible shortly.

PR #433, opened 2018-04-17
Jack properties use KSP_PIN, not KS_PIN.

timrprobocom:
The documentation for the four properties in KSPROPSETID_Jack all show that the input is a KS_PIN.  There is no such structure.  It should actually be KSP_PIN.  I added a link to the KSP_PIN page, and set the names to &gt;strong&lt; to match the other pages in this directory.


DOMARS:
@timrprobocom - Thanks Tim. Double checking with a few folks on KSP_PIN, but it certainly looks like KS_PIN isn't right.

DOMARS:
@timrprobocom - I checked with the dev team and they confirmed it should indeed be KSP_PIN. Thanks for an accurate PR to improve the Windows driver documentation.

PR #449, opened 2018-04-24
Update driver-flighting.md

SillyKeith:
Clarification failure reasons for CHID to be more clear.


aahill:
Thanks @SillyKeith, I'll merge this in and have it visible soon. Thanks for the contribution!

PR #451, opened 2018-04-24
Correct some misleading language.

timrprobocom:
The situation with driver signing is confusing enough, but there are things in this article that are misleading or inaccurate.  This makes some substantive changes, so you're probably going to want to review it for philosophy.


tedhudek:
Reviewing this internally, Tim.  Apologies for the delay.  We try to make it faster, but given the impact of this one I want to make sure we take a close look.  Appreciate the PR!

tedhudek:
Thanks again for the PR @timrprobocom. I just published an update to https://docs.microsoft.com/windows-hardware/drivers/install/kernel-mode-code-signing-policy--windows-vista-and-later- that incorporates your updates. I made a few updates on top after circling back with the team.  Note that you are now listed as a contributor on the byline of the page :)

PR #459, opened 2018-04-30
Fixed <a name=""> tag that showed as header

joshuaalm:
Removed double pound-sign that made anchor tag into h2 when rendered.  References to the anchor should still work.


tedhudek:
Thanks @joshuaalm for bringing this to my attention! I just removed about 2900 such occurrences. These were old anchors that were no longer necessary, so removal was the way to go. Closing out this PR accordingly.

PR #519, opened 2018-05-24
Update friendly-names-for-audio-endpoint-devices.md to match WCOS-related changes

JoelCorley:
- Fix a number of inaccuracies.  
- Add a description of how friendly names are discovered and how new device-specific friendly names should be based off of the device software key (HKR\MediaCategories).  
- Expand on the INF example.
- Eliminate the suggestion of making the GUIDs sequential as this can promote GUID collisions and reduces the amount of entropy in the GUID assignments.


DOMARS:
I was waiting to make sure the supported code changes were shipped and based on average branch velocity that should have definitely happened by now. 

@tedhudek  - This is approved to go live.

DOMARS:
Looks like there was an issue with an en-us in one of the links. I believe I just fixed that.

PR #526, opened 2018-05-24
Fixed link

EliotSeattle:
link to Memory allocation and Buffer Management
Now points to https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/content/_kernel/#memory-allocation-and-buffer-management


EliotSeattle:
This is in response to Issue 525

PR #539, opened 2018-06-02
Layout update (see description)

skycommand:
Level 3 headings were converted to level 2, because there was no level 2 in the document at all. Also, removed the hand-written table of content, because docs.microsoft.com automatically generates a TOC from level 2 headings. Finally, moved the redundant and technically incorrect repetition in the headings (which was inconsistent with the manual TOC too) to a sentence in the lead section.


aahill:
Thanks @skycommand, I'll merge and promote this live. Thanks for the pull requests! We certainly appreciate them!

PR #541, opened 2018-06-02
Layout and TOC fixes #2

skycommand:
Removed one empty level 2 heading, which caused an empty TOC in docs.microsoft.com. (I also deleted its anchor, but that should be fine, given that the section is at the top of the page anyway.) Converted Level 3 headings to level 2, because there was no level 2 in the document at all. The automatically generated TOC would pick them up, so it is a plus.


aahill:
Thanks @skycommand, this change should be up shortly!

PR #545, opened 2018-06-04
Improve InfVerif documentation

zlockard:


tedhudek:
Pushed commits we don't want here.  Will reopen in private mirror

PR #547, opened 2018-06-05
Integrate updates from IHV feeback

omarcardona:


duncanmacmichael:
Thanks for the PR, @omarcardona . Looks good; I'll merge it so the changes should be live later today.

PR #567, opened 2018-06-11
Clarify WPP providers limitation

Shyamal-V:


tedhudek:
Thanks, this will be live in 5

PR #570, opened 2018-06-12
Clerify what in the INF and code example the developer needs to change

SoundersFan:
This will hopefully avoid developers only changing the capability ID and not changing the interface GUID.


tedhudek:
Thanks @SoundersFan this will be on the live site later today.

PR #573, opened 2018-06-12
Clarify VS versions

ddv239:


ddv239:
It looks like my understanding of the possible version combinations between WDK and VS was simpler than it actually is. I am going to experiment with installing them and seeing what works.

ddv239:
Yes, I can definitely confirm that VS 2015 does have the Windows debuggers integration. So my change should be adjusted to accurately reflect the situation.

PR #577, opened 2018-06-13
Clarify supported WDK versions

ddv239:


tedhudek:
Thanks @ddv239.  @DOMARS  - so then the question would be, what happened in WDK 10 for debuggers?

ddv239:
I have made wrong conclusions from certain bug resolutions and assumed the WinDbg transport was removed from VS 2017. In reality, it is not, and for instance, having the RS3 WDK installed together with VS 2017 makes it successfully show up in VS. So this change is probably not factually correct.

PR #578, opened 2018-06-13
typo "your" should be "my"

SoundersFan:
make it better


tedhudek:
Thanks @SoundersFan this will be live soon

PR #582, opened 2018-06-15
Update get-product-data.md

balapv:
Corrected errors in OS Codes and OS Family Codes


aahill:
Thanks @balapv, I'll get this live soon!

PR #598, opened 2018-06-21
Adding 'System' to the registry paths.

PawelWMS:
The registry paths didn't include the "System" key and it was confusing for the users when they tried to copy-paste the path.


PageWriter-MSFT:
@PawelWMS. Yes that would be very confusing. Thanks for fixing the path.


Page Writer

PawelWMS:
@PageWriter-MSFT: You're welcome but I think you only closed the pull request without merging the changes. Are you going to do it in a separate commit?

PR #602, opened 2018-06-25
Layout fixes

skycommand:
Removed unused visible empty anchor. Applied standard list formatting to what was meant to be a list. Removed full stop from the end fragments; the full stop makes them look like imperative sentences.


barrygolden:
@skycommand thanks for the PR, will be live soon.

PR #610, opened 2018-06-27
add fMinimizeConnections

hankoh:
The policy was not specified.


duncanmacmichael:
Hi @hankoh , congrats on your first PR into the driver docs repo! Thanks for pointing out the name of the policy. I moved the name from the section header to the body text just to be consistent with the other section headers, but otherwise I'll go ahead and merge this PR for you. It should be live later today.

PR #622, opened 2018-07-02
Update attestation-signing-a-kernel-driver-for-public-release.md

SillyKeith:
Updated the page to call out which PE Levels can and cannot be achieved through attestation.   Added the CANNOTs to the NOTES at the top and the CANs in the section about acceptable PE binaries/levels.


aahill:
Thanks @SillyKeith, awesome PR as always! I'll get this up shortly.

PR #625, opened 2018-07-03
Fixed a huge visual bug

skycommand:
Fixed a huge rendering problem. I'm including a screenshot to show you what exactly is fixed. They include:
- Removal of two quote boxes. This means I removed one `>` from every single line.
- Changing `\[!Caution\]` to `[!Caution]` because the former is definitely what was intended.

The diff view in GitHub isn't impressive. You might want to use a different tool like Meld to compare the changes and generate a better visual diff.

screenshot_2018-07-03 bcdedit set


barrygolden:
@skycommand thanks for fixing these issues.

skycommand:
Glad to be of help! 😊

PR #627, opened 2018-07-03
Update windows-certified-products-list.md

SillyKeith:
Sysdev has retired and is offline.  so I removed the links to Sysdev and put a notice saying the feature will return in Fall 2018


aahill:
Hi @SillyKeith 

thanks for the PR! Looks good as always. I'll merge this in and have it live shortly.

Thanks,
Aaron

PR #663, opened 2018-07-16
Update whql-release-signature.md

Cloud-Writer:
Updated the existing note to avoid customers calling Support for clarification of this info.


tedhudek:
Thanks for the PR @Cloud-Writer this should be live in a few minutes

PR #668, opened 2018-07-17
Adding a link for a mandatory interface for device MFT

jiteshkris:
Device MFT also needs the IMFMediaEventGenerator along with the other listed. Adding the interface and the link for the interface


barrygolden:
@jiteshkris thanks for the PR. Your changes should be live soon.

PR #674, opened 2018-07-19
tweaks to enhance dch audio driver doc

nehinnan:
Adding clarity


tedhudek:
Thanks so much @nehinnan for the PR, and apologies for the huge delay. This PR for some reason became inaccessible due to a bug in the GitHub platform, and only just became accessible again today after I had an email thread with GitHub support!  Changes will be live within a few minutes.

PR #679, opened 2018-07-22
Deleted one dead link and the advice to use it

skycommand:
Deleted one dead link and the corresponding advice to use it


barrygolden:
@skycommand and @DOMARS, merged the PR, should be live later today.

PR #691, opened 2018-07-24
Minor layout fix

skycommand:
The "Symbol package deprecation" section was L3 instead of L2. As such it didn't appear in the TOC. Also, I formatted a list so that it looks like ... a list.


skycommand:
@DOMARS Done.

PR #701, opened 2018-07-26
Update how-do-i-generate-trace-messages-from-a-header-file-.md

sarman1998:
I was just trying to get this working and had a hard time until I realized that I needed to pass the .h file names i wanted processed on the tracewpp command line. and also that i couldn't use *.h because my trace config file trace.h was being included and tracewpp didn't like that either.


duncanmacmichael:
Hi @sarman1998 , thanks for the PR and congrats on your first contribution to the Microsoft driver docs! I polished some formatting for you, but otherwise this looks good so I'll merge this for you. It should be live soon.

PR #706, opened 2018-07-27
Update mobile-broadband-implementation-guidelines-for-usb-devices.md

jefsong:
Updated the expected behavior when network is assigning DNS to the host in the traditional manner.


duncanmacmichael:
Hi @jefsong , thanks for the update! I'll merge this for you and it should be live later today.

PR #724, opened 2018-08-02

NabilFates:
Documenting the flags for AddSoftware. Please do not publish this until we have internal signoff on the wordage.


EliotSeattle:
@tedhudek @duncanmacmichael @barrygolden 
Making sure you see this:
**Please do not publish this until we have internal signoff on the wordage.**

duncanmacmichael:
Yep, got it. I'll publish when the word is given. :)

PR #727, opened 2018-08-03
Added exception code

LuigiBruno:
Added an entry and description for the 0xC0000005 exception code.


DOMARS:
@LuigiBruno - Sorry for the delay on this PR. As you also submitted a related issue, I ended up taking the information in this PR an including it in other updates. The topic now includes a link to the NTSTATUS topic and a pointer to the header in the inc directory. I also mentioned the STATUS\_ACCESS\_VIOLATION as one example. Thanks for your interest in our docs and the ideas contained in this PR.

PR #738, opened 2018-08-08
Fixed case of WdfSpinLockRelease

ajbarb:


duncanmacmichael:
Hi @ajbarb , thanks for the PR! Good catch on the spelling case. I'll merge this PR for you and it should be live later today.

PR #761, opened 2018-08-15
Update hardware-dashboard-faq.md

SillyKeith:
removing the en-us from the support link to be more language agnostic and adding the new name for the final issue type selection choice.


EliotSeattle:
Fixed link

PR #787, opened 2018-08-24
Update system-defined-device-setup-classes-available-to-vendors.md

aholeinmyhead:
GPS device class is not defined in the OS itself and as a result, INFs cannot use the class without a ClassInstall32 section.  A universal driver cannot have a ClassInstall32 section (and even if they did, we don't want them to ClassInstall32 a system defined class).  This entry is causing confusion, so it should be removed for now


duncanmacmichael:
Hi @aholeinmyhead , thanks for clarifying this. The changes look good, so I'll go ahead and merge this PR for you. It should be live later today.

PR #793, opened 2018-08-28
Update hardware-support-app--hsa--steps-for-driver-developers.md

smadasam:
explained the RS5 AllowAny option, and stated when multiple SCCD files started to be allowed (starting RS4) 

also, titles are capitialized inconsistently, most aren't capitalized, but some are FYI


tedhudek:
Thanks @smadasam sorry for the delay on this one.  Should be live in a couple minutes.  I did some minor editing.

PR #820, opened 2018-09-04
Fixed an issue I introduced in my original description

JocelynBerrendonner:
I mistakenly inverted the logic when writing the description of the property. PKEY_SensorData_IsValid should be set to TRUE when the sensor sample is valid.


mimisasouvanh:
Looks good! Thanks for catching this.

PR #821, opened 2018-09-04
Fix broken links in nfc-class-extension-.md.

cwize1:


tedhudek:
Thanks @cwize1 these will be live within the hour

PR #826, opened 2018-09-06

SillyKeith:
Made some changes to Extended drivers and added an example.


tedhudek:
Thanks for the PR @SillyKeith, really appreciate it.  This will be live within a few minutes.

PR #841, opened 2018-09-14
Couple of fixes

stevewhims:
Fixing invalid quote character, and adding missing info (the xml namespace declaration for the uap4 prefix).


tedhudek:
Thanks @stevewhims for the PR!

PR #853, opened 2018-09-24
Update setting-up-a-connection-to-a-virtual-machine-in-visual-studio.md

YouRancestor:
Visual Studio should be run as administrator when starting the debugging session.

### DIFF
--- a/windows-driver-docs-pr/debugger/setting-up-a-connection-to-a-virtual-machine-in-visual-studio.md
+++ b/windows-driver-docs-pr/debugger/setting-up-a-connection-to-a-virtual-machine-in-visual-studio.md
@@ -60,7 +60,7 @@ The host computer can be the same physical computer that is running the virtual 
 ## <span id="Starting_the_Debugging_Session"></span><span id="starting_the_debugging_session"></span><span id="STARTING_THE_DEBUGGING_SESSION"></span>Starting the Debugging Session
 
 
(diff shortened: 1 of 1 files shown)


DOMARS:
Duncan, in your experience would you launch VS as admin to debug when working back in the older versions of Windows that supported this?

duncanmacmichael:
When I was setting up a connection to a virtual machine in VS, I don't recall having to run VS as administrator, no. At the time I was using VS 2017 Enterprise, Windows 10, version 1709, and the 1709 version of the WDK. It could be different for older versions of Windows or VS but I can't say with 100% certainty. Perhaps we could add running as admin as an optional tip if someone encounters an error while starting the debugging session?

YouRancestor:
@DOMARS 
I received this error from Visual Studio:

> Windows Debugging Extension for Visual Studio
Could not start debug session, error 80070005: Access denined.

With Visual Studio Community 2017 15.7.3, Windows 10 Professional 64bit (10.0, version 17134), Windows Driver Kit - Windows 10.0.17134.1, and the virtual machine is the default Windows 10 dev environment(Enterprise Evaluation) provided by Hyper-V.

DOMARS:
@YouRancestor Thanks for this input. After looking at this, I ended up adding a tip relating to running as admin. That change is now live in the topic. 
You should be aware that you are running a version of the WDK that is newer then the last supported version (1507) for this functionality. 
Good luck with your debugging and thanks for this pull request.

PR #888, opened 2018-10-09
Additional steps in the "Resolution" section

LuigiBruno:
Added additional suggestion in the "Resolution" section after the first step (executing the "!analyze" debug extension) mentioning the "!thread" extension and the "dds,dps,dqs" commands to get additional information about the thread and the contents of the memory at the bug check time.


DOMARS:
@LuigiBruno - It would be great if you can fix the links. To allow for the best global experience, when we include links we remove the language specifier in the URL. Our web sites, detect the language in the browser and direct to that language experience.

So this would be the link we would use in topics, for example:

https://docs.microsoft.com/windows-hardware/drivers/debugger/-analyze

LuigiBruno:
Thanks @DOMARS.
I've just fixed the links and committed this new change

DOMARS:
@LuigiBruno - Thanks for this PR. I'm looking at this stop code to see if there is additional information to add here as well as your suggested steps / debugger commands.

PR #913, opened 2018-10-11
typo: Double word "down"

nschonni:


EliotSeattle:
I don't see the change for "down" but 
But these devices need a warning that the power status **in** changing 
should be 
But these devices need a warning that the power status **is** changing

EliotSeattle:
Thanks a million, @nschonni!!!
Did you use a script to go over the driver repo content? This is very helpful. VERY helpful. I appreciate the effort. We'll merge these changes today.
Just an FYI, you can add multiple commits in a single PR if that's easier for you. 
Again, THANK YOU!

nschonni:
@EliotSeattle using https://github.com/errata-ai/vale
You can suggest internally that they re-look at https://github.com/Microsoft/vscode-docs-authoring/issues/54 if you think it makes sense for the authors to see it in VSCode too

PR #923, opened 2018-10-12
typo: Double word "driver"

nschonni:


duncanmacmichael:
Thanks for the fix, @nschonni ! I've merged this for you and it should be live later today.

PR #944, opened 2018-10-15
Added RS5 EMI V2 support note

jatru-msft:
In Windows 10 RS5, devices can also support EMI_VERSION_V2.


DOMARS:
Thanks Jake, we will publish this update within 24 hours.

PR #953, opened 2018-10-18
Update introduction-to-adapter-objects.md

kvndb:
enhanced the paragraph


tedhudek:
Hi @kvndb thanks for the suggestion, but I think the paragraph is ok as-is.

PR #978, opened 2018-10-28
chore: Remove en-us from docs.microsoft.com links

nschonni:


mimisasouvanh:
The commit validation failure is a false-positive. The link is normal, and does not specify a locale.

PR #984, opened 2018-10-30
Update publishing-for-test-distribution.md

SillyKeith:
Removed the 14 day auto expiration wording and replaced with you must expire after finishing testings


tedhudek:
Thanks for the PR @SillyKeith this will be live in 15 minutes or so

PR #1023, opened 2018-11-08
Update elam-prerequisites.md

nbassett-MSFT:
Removing the reference to VIA and redirecting interested non-MVI partners to our alias directly.


tedhudek:
Thanks for the PR @nbassett-MSFT this should be live in a few minutes

PR #1024, opened 2018-11-08
Update winacpi-osi.md

AaDake:
Added "Windows 10, version 1809" in the "Applies to" section


tedhudek:
Thanks for the PR @AaDake this will be live in a few minutes

PR #1025, opened 2018-11-08
Removed UcmUcsi.sys references

radutta99:


tedhudek:
Thanks for the sharp eye @radutta99

PR #1047, opened 2018-11-14

AaDake:
Corrected the code language for the samples (ASL, instead of CPP)


EliotSeattle:
Looks good. Ship it!

PR #1054, opened 2018-11-16
Update device-metadata.md

SillyKeith:
All the links on this page were broken and redirected to devcenter main page.  Fixed them


EliotSeattle:
Merged changes.
FYI, Keith. When creating links, remove the /en-us part of the path. Otherwise, it forces the browser chrome into English, no matter what the locale of the customer.

PR #1103, opened 2018-11-29
Update DIFx guidelines

jasonknichel:


tedhudek:
Thanks for the PR @jasonknichel looks good.  Will be live in a couple minutes

Issue #1097, opened 2018-11-29
Unclear/inconsistent documentation of network device Surprise removal


I-gor-C:
From the doc:
> 5. If the miniport driver was successfully initialized, NDIS calls the MiniportDevicePnPEventNotify function with an event code of NdisDevicePnPEventSurpriseRemoved. The miniport driver should note that the device has been physically removed. If the miniport driver is an NDIS-WDM driver, it should cancel any pending IRPs that it sent down to the underlying bus driver. If the miniport driver was not successfully initialized, processing continues.
> 6. NDIS performs the following steps:
>     a. It pauses all protocol drivers that are bound to the NIC.
>     b. It pauses all filter drivers that are attached to the NIC.
>     c. It pauses the miniport driver for the NIC.

I suppose the _**processing continues**_ in par.5 means that the algorithm proceeds to the next step (6)? If that's true then 6.c describes an invalid transition in miniport SM. If initialization has failed the miniport returns to the Halted state. Triggering Pause from Halted state isn't a valid transition (per https://docs.microsoft.com/en-us/windows-hardware/drivers/network/adapter-states-of-a-miniport-driver ).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 619ff26a-1b82-74fe-db0a-b0b17e370730
* Version Independent ID: ba894464-1bd8-7d9e-6f7c-b6ccd21a4d9f
* Content: [Processing the surprise removal of a NIC (Windows 7 and later) - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/network/processing-the-surprise-removal-of-a-nic--windows-7-and-later-versions-)
* Content Source: [windows-driver-docs-pr/network/processing-the-surprise-removal-of-a-nic--windows-7-and-later-versions-.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/network/processing-the-surprise-removal-of-a-nic--windows-7-and-later-versions-.md)
* Product: **windows-hardware**
* GitHub Login: @duncanmacmichael
* Microsoft Alias: **dumacmic**

duncanmacmichael:
Hmm, interesting question @I-gor-C. I'm not sure, either, so let me see what I can find out from the feature team to clarify what the doc is saying.

jtippet:
[...]

PR #1109, opened 2018-11-30
processing-the-surprise-removal-of-a-nic--windows-7-and-later-versions

jtippet:
Clarify the versioning and step 5.

Addresses #1097


duncanmacmichael:
Looks great @jtippet, thanks for the clarification! This should be live later today.

PR #1124, opened 2018-12-04
Remove absolute registry path

jasonknichel:
Remove absolute registry path and link to the correct way of determining the relevant settings location.


duncanmacmichael:
Hi @jasonknichel, thanks for the PR. This looks good, so I'll go ahead and merge it for you. The change should be live later today.

PR #1130, opened 2018-12-09
Fix !positions command

felipepessoto:


tedhudek:
Thanks @felipepessoto this will be live soon

felipepessoto:
@tedhudek, I think there is a bug in the docs.microsoft.com integration with GitHub, or maybe a stale cache. The link in my name points to my old username (I changed months ago):

image

Do you know how to fix it? Thanks

tedhudek:
@felipepessoto so the commit on that file is 43b2c6388 which has your hotmail as the author email. Is your new GitHub account connected to a different email address?

tedhudek:
I understand. But the system uses email address as the key.

felipepessoto:
Thanks @tedhudek, if the component is open source I'll be glad to help find the bug.

PS: I didn't delete/recreated the account, I renamed it

PR #1132, opened 2018-12-10
Update validating-universal-drivers.md

Cymoki:
Updating doc with known issues.


tedhudek:
Thanks for the PR @Cymoki

PR #1144, opened 2018-12-15
ETW tools documentation updates

Robo210:
- The TraceView, tracelog, tracepdb tools documentation is out of date.
  - Remove mention of OS versions. None of these tools work on Windows 200 anymore; they all require Windows 7 or later.
  - Remove mention of unsupported features, such as TraceView's command line support (use tracelog instead) and Windows 2000-specific tracing.
  - Replace uses of the word "Crimson" with more specific descriptions of what is being referenced. "Crimson" was an internal code word for the Windows Vista rewrite of ETW and the Event Log.
  - Replace \*\*Note\*\* with DocFx Notes.
- Remove a lot of excess whitespace that markdownlint was upset about.
- Try to clarify some WPP-specific ETW behaviors that don't apply anymore to other forms of ETW logging.
  - There is still a lot of work to be done in this regard. It would probably be best to throw away a lot of these ETW pages and just direct viewers to the far-more-complete SDK documentation.


windows-driver-docs-pr/devtest/control-guid-file.md
```diff
@@ -26,7 +24,7 @@ Developers of trace providers typically supply this file. However, if you have t
 
 In the source code, find the definition of the [WPP\_CONTROL\_GUIDS](https://msdn.microsoft.com/library/windows/hardware/ff556186) macro. The GUID value and GUID friendly name are shown in bold in the example below.
 
-```
+```C
```
idigdoug: Intended?

Robo210: It specifies the language as C, and enables syntax highlighting.

PR #1181, opened 2018-12-27
Fix misplaced code fences in windows-driver-docs-pr/wdf/using-device-interfaces.md

cxw42:
In the code example, all of item 4 was rendering in a code block.  This change corrects the positioning of the fences so that is no longer the case.


tedhudek:
Thanks @cxw42 good find!

PR #1193, opened 2019-01-04
Update guidelines-for-location-simulators.md

Cheoljin:
Add clarification for the simulator.


tedhudek:
Thanks for adding this point @Cheoljin, this doc will be live momentarily.

PR #1203, opened 2019-01-08
Edited the bug check's description.

LuigiBruno:
Provided an additional information about the meaning of this bug check error.


DOMARS:
@LuigiBruno - Thanks for this PR. Because we have not documented information on this bug check before, I would need to check with the dev team to be able to do that. As I have mentioned before, we use a process to prioritize the most frequent bug checks out of the hundreds that are currently defined. If you think this is a popular bug check, you could open an issue for this topic with that request. With our limited documentation team resources, it is likely that we will only be able to do the work to create and validate a universal description for a small sub set of all of the possible stop codes. In my limited for this stop code, it does not to be one that occurs frequently.

PR #1206, opened 2019-01-08
Description details

LuigiBruno:
Added further information in the bug check's description.

### DIFF
--- a/windows-driver-docs-pr/debugger/bug-check-0x3--invalid-affinity-set.md
+++ b/windows-driver-docs-pr/debugger/bug-check-0x3--invalid-affinity-set.md
@@ -18,7 +18,7 @@ ms.localizationpriority: medium
 
 The INVALID\_AFFINITY\_SET bug check has a value of 0x00000003.
 
-This bug check appears very infrequently.
+This bug check appears very infrequently. It indicates a null of incorrect subset affinity.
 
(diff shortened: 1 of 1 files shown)


windows-driver-docs-pr/debugger/bug-check-0x3--invalid-affinity-set.md
```diff
@@ -18,7 +18,7 @@ ms.localizationpriority: medium
 
 The INVALID\_AFFINITY\_SET bug check has a value of 0x00000003.
 
-This bug check appears very infrequently.
+This bug check appears very infrequently. It indicates a null of incorrect subset affinity.
```
tedhudek: Not sure I understand what this means. Can you be more specific @LuigiBruno?

DOMARS:
@LuigiBruno - Thanks for this PR. Because we have not documented information on this bug check before, I would need to check with the dev team to be able to do that. As I have mentioned before, we use a process to prioritize the most frequent bug checks out of the hundreds that are currently defined. If you think this is a popular bug check, you could open an issue for this topic with that request. With our limited documentation team resources, it is likely that we will only be able to do the work to create and validate a universal description for a small sub set of all of the possible stop codes. In my limited for this stop code, it does not to be one that occurs frequently.

PR #1209, opened 2019-01-08
Description and resolution

LuigiBruno:
Added further information in the bug check's description.
Added the "Resolution" section to suggest very first analysis action to be done.


tedhudek:
Thanks for these fixes @LuigiBruno

PR #1216, opened 2019-01-11
We're no longer calling it the Slate Audio sample.

girishpattabiraman:


DOMARS:
@tedhudek - This PR is ready to go live.

PR #1253, opened 2019-01-22
Corrected a broken link to CheckINF (now InfVerif).

heinsaar:


EliotSeattle:
Thanks, @heinsaar for the correct link.

PR #1259, opened 2019-01-23
Corrected a wrong description.

heinsaar:


tedhudek:
Thanks @heinsaar and apologies for the delay.  I just merged this fix in the private mirror, should be live within 15 min or so

PR #1264, opened 2019-01-24
Updated description

LuigiBruno:
Added information to the bugcheck error description.


DOMARS:
@LuigiBruno - Thanks for this PR. There are times where the bugcheck may not follow the pattern that the name would suggest. So for this reason to add any description I need to research the code and work with the developer team. As I have mentioned before, with our limited documentation team resources, it is likely that we will only be able to do this work for a small sub set of all of the possible stop codes.

PR #1292, opened 2019-02-01
Update SDV page to include installation section 

NateD-MSFT:
SDV dev here.  I'm adding a paragraph describing where SDV can be obtained and what the prerequisite Visual C++ redistributables are to use it.


EliotSeattle:
Looks fine. Thanks, @NateD-MSFT

PR #1312, opened 2019-02-08
Edited bug check description

LuigiBruno:
Edited the bug check description.


LuigiBruno:
Thanks for the feedback, @DOMARS.

PR #1320, opened 2019-02-12
Update attestation-signing-a-kernel-driver-for-public-release.md

SillyKeith:
replacing CABs are not required to they are required.


EliotSeattle:
Sorry, @SillyKeith. This got lost in the snow for a couple of days. It's now on the review site and should be live by morning.

PR #1323, opened 2019-02-14
Add a note about a VPN adapter

aktsuda:
Add a note about a VPN adapter as follows:
**Note** A VPN adapter should set **\*NdisDeviceType** to 0. If it set **\*NdisDeviceType** to NDIS\_DEVICE\_TYPE\_ENDPOINT (1), a UWP is prevented from sending a packet to the VPN adapter.


duncanmacmichael:
Hi @aktsuda, thanks for the PR. I made some minor editorial polishes, but otherwise the change looks good so I'll get this merged for you. It should be live later today, Redmond time. Thanks!

PR #1326, opened 2019-02-14
Update update-a-code-signing-certificate.md

SillyKeith:
Updated page to include instructions for adding "additional certificates" as the process is the same.


EliotSeattle:
Thanks, @SillyKeith

PR #1333, opened 2019-02-16
Update DVL documentation to reflect SDV/CA order.

NateD-MSFT:
SDV dev here.  SDV performs a clean rebuild of drivers, which removes the Code Analysis log as a result.  This change specifies that SDV must be run *before* CA to keep both logs for generating a DVL.

Also removing some outdated language that refers to Win8/Server 2012.


DOMARS:
@tedhudek - This PR from Nate looks good to me.

PR #1337, opened 2019-02-19
Update installing-preview-versions-wdk.md

Cymoki:
Updating to point to VS2019 which is the new requirement for the next WDK.


Cymoki:
ready to push

PR #1353, opened 2019-02-26

SillyKeith:
Added the limitation about including any buildnumber in the Manufacture section - Please reveiw that my hyperlink back to that section conforms and will work when public.


EliotSeattle:
Looks good.
[**INF Manufacturer section**](../install/inf-manufacturer-section.md) works just fine in preview. For links to other folders, you can also just use the full URL of the topic, minus the languageID, en-us, say.

PR #1367, opened 2019-03-01
Can't Assume IRQL PASSIVE_LEVEL

PeterGV:
Please don't encourage people to assume the IRQL will be PASSIVE_LEVEL, as the parenthetical in paragraph at line 239 does now.  Either remove the parenthetical entirely (this is the best idea) or at the warning  that the driver code must never RELY on the IRQL being PASSIVE_LEVEL in this case.  The only time a driver has the "right" to assume the IRQL is PASSIVE_LEVEL in an EvtIo Event Processing Callback is when the driver has established the appropriate ExecutionLevel constraint.  No other time.  Anything else requires people to rely upon an artifact of implementation (that the user-mode caller's IRQL is passed-through by the Framework), which is a mistake.


tedhudek:
Thanks for the suggestion @PeterGV. Would you be ok removing the parenthetical entirely, as you sugges above? I note in your PR you add a qualifier to it.

PeterGV:
Hi Ted!

I would actually prefer to remove the parenthetical entirely. But I’d didn’t want to presume to do this in the PR.

If that’s good for you, that’s definitely good for me.

P

—-
From mobile
________________________________

PR #1393, opened 2019-03-11
Update getting-started-with-universal-drivers.md

noishiza:
I request to update a document about universal windows driver because it is not specified the target device type in the document. This guidance is for in-box driver, not for add-on device driver.


tedhudek:
Hi @noishiza thanks for the PR.  I think that this guidance applies to all 3rd party drivers, not just inbox.  Can you provide more info on why you think this is for inbox drivers only?

Adding @selerner fyi

noishiza:
Hi Ted,
Sorry, please pend this request.
I am confirming this to PG now.

Regards,
Nozomu

tedhudek:
Hi @noishiza checking back on this one.  Do you have any more info?

noishiza:
Hi @tedhudek, I withdraw this request. I'm sorry to trouble you.

tedhudek:
No problem at all @noishiza!  Please let us know if we can help in the future.

PR #1407, opened 2019-03-14
Fixing instructions for loading symbols for MyApp

msatranjr:
.symfix will reset sympath to Microsoft public symbol server and since we are debugging on the same machine that we compiled the code, setting .srcpath is unnecessary. Symbols already have hard-coded paths to source files, which will work automatically because the target and the machine that compiled the code are the same. Removing this step will not only reduce confusion, but it will reduce the potential for mistakes.


msatranjr:
.symfix will reset sympath to Microsoft public symbol server and since we are debugging on the same machine that we compiled the code, setting .srcpath is unnecessary. Symbols already have hard-coded paths to source files, which will work automatically because the target and the machine that compiled the code are the same. Removing this step will not only reduce confusion, but it will reduce the potential for mistakes.

DOMARS:
@msatranjr - Thanks (xiè xiè) for the interest here and the PR. 
I'm going to have to think about this as the benefit of setting a path to the symbol server is public symbols for other things beyond your code. So I think I would agree with what you are saying for the symbols for your code, but often times having other symbols can be useful for figuring out why your code is not working.

DOMARS:
@tedhudek - This PR is approved and ready to go to live/master when you have a minute.

PR #1432, opened 2019-03-28
Make WdfDeviceInitSetReleaseHardwareOrderOnFailure usage more discove…

Shyamal-V:
…rable for bus drivers

Make WdfDeviceInitSetReleaseHardwareOrderOnFailure usage more discoverable for bus drivers by calling it out in the power up/down sequence


tedhudek:
Thanks @Shyamal-V!  Should be live now

PR #1447, opened 2019-04-05
Update attaching-a-debugger-to-the-print-filter-pipeline-service.md

hsebs:
'1' is missing in step 3.
90000 is '1.5 minutes', it is not '.5 minutes'


barrygolden:
Thanks for the PR, @hsebs!

PR #1449, opened 2019-04-08
adding MSDL service status + contact info

herenhuang:
The goal here is to have a "status box" on this page to display to customers whether the public symbols server is up or not + an official avenue to report issues, which we currently don't have. The status will be edited as part of any livesite work if issues come in so will require a manual edit from our on-call folks. Historically this has only happened once every two months or so. The next step is to see if we can do any automated azure monitor piping so the status can be automatically updated. I just realized I didn't edit this with a corporate GitHub account so please let me know if I need to do so (hehua).


DOMARS:
@barrygolden - This PR is ready to go when you have a moment.

PR #1450, opened 2019-04-08
removing a duplicate entry

herenhuang:
removing the symstore link cause it showed up twice


DOMARS:
@barrygolden - This PR is ready as well.

PR #1459, opened 2019-04-12
Fix Highlightighting typo

thecodetinker:
I assume it was a typo?!


EliotSeattle:
Oops, acted on the Issue before I saaw the PR. Thanks, @thecodetinker.

PR #1475, opened 2019-04-18
Update download-the-wdk.md

Cymoki:
correcting the download link for WDK.


EliotSeattle:
Content validation failed, so I am making the change through another PR.

PR #1506, opened 2019-04-30
Update register-for-the-hardware-program.md

SillyKeith:
Adding a Support callout to this section as they will not have access to Dev Center to be redirected.


EliotSeattle:
POST BUILD. Approved for publish.

PR #1508, opened 2019-04-30
Permit non-DX APIs to sever WDK dependencies.

BKlamik-MSFT:
@jjulianoatnv: Is this sufficient?


EliotSeattle:
Content is frozen until Build on 5/6. Will release the changes then. Thanks!

PR #1538, opened 2019-05-07
Add details about driver inf modifications.

BKlamik-MSFT:


EliotSeattle:
@BKlamik-MSFT 
Did you want the code entries to show as code?

PR #1539, opened 2019-05-08
Update inf2cat.md to include the new and missing OS Codes

rachand-hdc:
New OS Codes are updated. Some missing OS codes are added.


EliotSeattle:
Thanks, @rachand-hdc!!

PR #1596, opened 2019-06-04

SillyKeith:
removed double "be"   "be be"


EliotSeattle:
2 Be or not 2 Be? ;-)

PR #1597, opened 2019-06-04
Removing device default behavior.

omarcardona:
Removing 0 enumeration which is vague and including guidance on transport selection and multi-transport behavior.


duncanmacmichael:
Looks good @omarcardona, thanks for the updates. I'll go ahead and merge this for you, and it should be live shortly.

PR #1635, opened 2019-06-24
Updating WinDbg What's New for 1905 release

aluhrs13:


aluhrs13:
@EliotSeattle can you help me out with this since Don is OOF?

PR #1641, opened 2019-06-26
Add proper documentation on "-pc"

TimMisiak:
The actual functionality of -pc is "pending create" which creates an implicit command line option that can be used via the -cimp option on the client side.


DOMARS:
@EliotSeattle - This PR is ready to go to master/live on Monday.

PR #1646, opened 2019-06-27
Update bug-check-code-reference2.md

nnpcYvIVl:
Fix incorrect names


DOMARS:
@EliotSeattle - These fixes to the display names for the linnks look good to take to Master/Live on Monday.

PR #1655, opened 2019-07-08
fix: remove unnecessary dbgcmd log

profnandaa:
The previous log included a log for a previously ran command,
which might be confusing to a beginner. Perhaps just starting
off from the command, as consistent with the rest will be ok.


DOMARS:
@profnandaa -Thanks for noticing this inconsistency and providing a PR. The hitting a breakpoint output should actually be in a code block right after the: "Notepad runs until it comes to the WinMain function, and then breaks in to the debugger." text. Would you be willing to update this PR to move this text in a code block with the dbgcmd attribute right after the breakpoint text?

profnandaa:
@DOMARS - sure, just done 👍

DOMARS:
@barrygolden - This PR is ready to go to live when you have a minute.

PR #1661, opened 2019-07-11
Make validation a subsection under Wake on Voice.

chcurlet-msft:
'Validation on Modern Standby Systems' should be a subsection under WoV since this refers to validating WoV specifically.


EliotSeattle:
Looks good. Thanks for the PR, @chcurlet-msft

PR #1664, opened 2019-07-15
Fixing paths

vivekspell:
Fixed html tags.


EliotSeattle:
@tedhudek, I was going to accept this PR, but shouldn't angle brackets in an HTML block **require** that they be escaped? Do we have a behavior that's different here?

tedhudek:
Staging looks good.  Thanks @vivekspell!

tedhudek:
Staging looks good.  Thanks @vivekspell!

PR #1672, opened 2019-07-19
Update text to avoid issue in localization

daxianji007:
The English page would not change.


DOMARS:
@EliotSeattle - Could you take a look at this and advise? The @ symbol seems to render OK as is. Would it be advisable to take this loc PR?

EliotSeattle:
Escaping the ampersand is completely fine, outside of code blocks. Looks good, @DOMARS

DOMARS:
@tedhudek - This PR is ready to go to master/live when you have a moment.

PR #1696, opened 2019-07-31
Adding notice of cross cert expirations

nasantia:
Adding notice box of cross cert expiration to relevant pages, as well as adding FAQ for users to navigate to to learn more about the situation


tedhudek:
Will reopen with latest updates in the private mirror

PR #1713, opened 2019-08-09
Fixed MUTT software package link

radutta99:
The link was a stale msdn page that went to default USB page.


duncanmacmichael:
Thanks @radutta99, I went ahead and removed the en-us specific locale from the URL but otherwise this looks good. I'll get this merged, and it should be live shortly.

PR #1725, opened 2019-08-19
Update create-a-new-hardware-logo-submission.md

SillyKeith:
HDC has a limitation of 160 characters for the relative path.  added this requirement.


SillyKeith:
abandon

PR #1778, opened 2019-09-19
Update heap() example output

emmasab:
Command example output for Heap() was the output of a different command, Calls()


DOMARS:
@tedhudek - This PR is ready to go to master/live when you have a minute.

PR #1785, opened 2019-09-20
Update ksproperty-cameracontrol-extended-vfr.md

gerritsMSFT:
The correct implied settings for frame rate should be variable. All current cameras without this control use variable by default, and lack of this control should not contradict that.


barrygolden:
Thanks for the PR @gerritsMSFT !

PR #1791, opened 2019-09-25
Remove/update some references to hardcoded registry locations

jasonknichel:


tedhudek:
Thanks @jasonknichel this looks great!  Should be live within 20 min or so

PR #1794, opened 2019-09-27
Update microsoft-public-symbols.md

anaik3:


DOMARS:
@tedhudek - This PR is ready to go to master/live.
(Looks like they were able to quickly fix the symbol server, so this text is to update symbol server status)

PR #1792, opened 2019-09-26
Document executable name option in WinDbgX cmd line

xiaoyinl:
`WinDbgX.exe [executable]`: launch the executable and debug it.
The description is copied from https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/windbg-command-line-options

PR #1806, opened 2019-10-07
Fix table format inconsistency introduced in #1792

xiaoyinl:
Add missing | in a table entry added in #1792. Although it's optional, I think it's better adding it for consistency with the code above.


DOMARS:
@tedhudek - This PR is ready to go to live/master when you have a moment.

PR #1810, opened 2019-10-11
Updating WinDbg Preview License to be less restrictive.

aluhrs13:
Our current EULA is pretty restrictive due to the "Preview" nature of it. This is less restrictive and essentially a "GA" license to match normal shipping apps. This is all from CELA, so I'd normally "fix" some of the diff to look cleaner, but I'd rather keep everything exactly as CELA gave it to me.


DOMARS:
@tedhudek - This PR is ready to head up to master/live when you have a moment.

PR #1836, opened 2019-10-23
Update windows-network-architecture-and-the-osi-model.md

xiaoyinl:
The full name of [ISO](https://www.iso.org) should be "the International Organization for Standardization".


duncanmacmichael:
Hi @xiaoyinl, nice catch! You're indeed correct, so I'll go ahead and get this PR merged for you so we can get the fix live within the next few minutes. Thanks!

PR #1838, opened 2019-10-23
Update process-syntax.md

BissetJ:
The example in the 'An example of the |[Expression] syntax ...' paragraph was malformed.  The pipe (|) should be outside the square brackets, not in.  That is, the example was supposed to be this:
|[@$t0]
not this:
[|@$t0]


DOMARS:
@tedhudek - This PR is ready to head up to Master/Live when you have a minute.

PR #1842, opened 2019-10-24
Clarify SDV library limitations

NateD-MSFT:
SDV owner here.  This page states that the class framework library limitations apply only to WDM drivers, which is incorrect.  They apply to all drivers; this new language reflects that.  This change also indicates that some checks can still be performed even if the driver does not adhere to these limitations.

This change also makes explicit that SDV can only process driver libraries, not generic C/C++ libraries.


DOMARS:
@tedhudek - This PR is ready to head up to Master/Live when you have a minute.

PR #1851, opened 2019-10-28
Update microsoft-public-symbols.md

Contagion21:


DOMARS:
@tedhudek - This PR is ready to go. This is a real time update to the symbols download page, It would be good to take this to master/live as soon as we can.

PR #1852, opened 2019-10-28
Fix command to enable testsigning

rahul-ramadas:
The original command might have specified `{default}` which someone
might have incorrectly assumed to be a placeholder for a GUID.


tedhudek:
Thanks @rahul-ramadas this fix should be on the live site within the hour.

PR #1896, opened 2019-11-11
removed the sentence because it appeared twice

RAJU2529:
i have removed the following sentence
You can also manually run the Stampinf tool, which is located in the bin subdirectory of the WDK.
As per user issue, I made this PR


tedhudek:
Thanks @RAJU2529

PR #1904, opened 2019-11-16
replaced wrong link to new link

RAJU2529:
As per the user issue # 1903
 i have replaced the wrong link
**https://docs.microsoft.com/en-in/windows-hardware/drivers/ddi/index**

to new correct link
**https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/bcd-boot-options-reference**

also added a related  topic link
**https://docs.microsoft.com/en-us/windows-hardware/manufacture/desktop/bcdedit-command-line-options**

### DIFF
--- a/windows-driver-docs-pr/devtest/editing-boot-options.md
+++ b/windows-driver-docs-pr/devtest/editing-boot-options.md
@@ -14,13 +14,9 @@ ms.localizationpriority: medium
 
 # Editing Boot Options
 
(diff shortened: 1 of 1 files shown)


RAJU2529:
any updates from this PR

EliotSeattle:
Thanks, @RAJU2529 for creating the pull request. I really appreciate your contribution.
One quick note for future reference: when creating a link using the fully qualified URL, remove the locale identifier, in this case, "en-us". If it's left in, we force the browser to the English version. If it's removed, the browser will request the customer's locale setting. If we have a page localized into that language, we'll serve it up, rather than forcing English.
Again, thanks so much for the PR!
Eliot

PR #1927, opened 2019-11-27
Corrected WER link in documentation body

ayodejiige:


DOMARS:
@ayodejiige - I'm following up on this to see if I can get information on the OCA program.

PR #1932, opened 2019-12-02
Update installing-windows-symbol-files.md

trappergor:
trunslate for russian


DOMARS:
@trappergor – Thanks for this contribution. Our current content system is not set up to take community contributions for translated text, so we are not able to approve this PR. We are considering adding the ability to support community translation, and we will take your interest in doing so into account. Thanks again for your interest in our docs.

PR #1937, opened 2019-12-03
Added "Resolution" section

LuigiBruno:
Added "Resolution" section with the first analysis action.


LuigiBruno:
> @LuigiBruno - Thanks for PR. BTW, some of these older bug checks are no longer active in current day Windows, but if anyone runs into it, using !analyze is always a good idea.

Thank you @DOMARS.

PR #1947, opened 2019-12-09
minor tweaks to description of standalone driver package

jasonknichel:


tedhudek:
Looks good, thanks for the tweak

PR #1981, opened 2020-01-10
Update understanding-and-configuring-windows-connection-manager.md

msarcletti:
This "Signal strength" section it true only for Windows 8 and 8.1 but NOT for Windows 10. I'd like to add this information


EliotSeattle:
Merged. Thanks!

PR #2013, opened 2020-01-27
fix dtrace.org link

Mixaill:
dtrace.org does not support HTTPS


EliotSeattle:
Thanks for catching this, @Mixaill

PR #2016, opened 2020-01-28
Typo

jozefizso:


EliotSeattle:
Thanks, @jozefizso 
Looks like we have a hard time with the big words. Appreciate the PR!

PR #2034, opened 2020-02-10
Clarity fixes for KDNET docs

aluhrs13:
Adding some clarity to disambiguate "kdnet" the protocol and "kdnet.exe" the tool, as well as cleaning up the directions on when to restart the target.


DOMARS:
Looking at this...

Issue #2038, opened 2020-02-12
INF Version section link is wrong


baget:
INF Version (in Remarks and also in See Also) section link is wrong, it point to INF AddService and not INF Version
it should point to
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-version-section

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cdb22e15-68dc-98c7-6ef9-eacc7657ad16
* Version Independent ID: 5d242ff3-a365-9ff4-12d7-4cf27a23c099
* Content: [DEVPKEY_Device_ClassGuid - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/devpkey-device-classguid#feedback)
* Content Source: [windows-driver-docs-pr/install/devpkey-device-classguid.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/install/devpkey-device-classguid.md)
* Product: **windows-hardware**
* Technology: **windows-devices**
* GitHub Login: @tedhudek
* Microsoft Alias: **tedhudek**

RAJU2529:
@illfated . please review this PR #2041

tedhudek:
Fixed in #2041

PR #2041, opened 2020-02-15
Drivers/install: INF Version link correction

RAJU2529:
as per the user report #2038 
i replaced old link to new link
**https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-version-section**

### DIFF
--- a/windows-driver-docs-pr/install/devpkey-device-classguid.md
+++ b/windows-driver-docs-pr/install/devpkey-device-classguid.md
@@ -83,14 +83,13 @@ Requirements
 ## See also
 
 
-[**INF Version Section**](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-version-section)
+[**INF Version Section**](https://docs.microsoft.com/windows-hardware/drivers/install/inf-version-section)
 
 [**SetupDiGetDeviceProperty**](https://docs.microsoft.com/windows/desktop/api/setupapi/nf-setupapi-setupdigetdevicepropertyw)
 
  
 
  
-
 
 
 
--- a/windows-driver-docs-pr/install/devpkey-device-classguid.md
+++ b/windows-driver-docs-pr/install/devpkey-device-classguid.md
@@ -83,7 +83,7 @@ Requirements
 ## See also
 
 
-[**INF Version Section**](https://docs.microsoft.com/windows-hardware/drivers/install/inf-addservice-directive)
+[**INF Version Section**](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-version-section)
 
 [**SetupDiGetDeviceProperty**](https://docs.microsoft.com/windows/desktop/api/setupapi/nf-setupapi-setupdigetdevicepropertyw)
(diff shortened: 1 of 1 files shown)


windows-driver-docs-pr/install/devpkey-device-classguid.md
```diff
@@ -83,7 +83,7 @@ Requirements
 ## See also
 
 
-[**INF Version Section**](https://docs.microsoft.com/windows-hardware/drivers/install/inf-addservice-directive)
+[**INF Version Section**](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-version-section)
```
illfated:
```suggestion
[**INF Version Section**](https://docs.microsoft.com/windows-hardware/drivers/install/inf-version-section)
```
Links should be kept language agnostic, especially if they were so before.

illfated:
Please make the same change (link file name) on line 21:
https://github.com/MicrosoftDocs/windows-driver-docs/pull/2041/files#diff-bcd4274f6733b04c57ac40f93bdf33dfL21
The same link is shown there too.

RAJU2529:
@illfated . Just now i entered below address  
**https://docs.microsoft.com/windows-hardware/drivers/install/inf-version-section**

and above url is  redirected to 

**https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-version-section**

so what should i do ? still you prefer without en-us  in url

illfated:
OK, I see. I did not know that you do not know about Microsoft URL language redirection and translated pages.

1. Some of the Microsoft Docs pages are translated to other languages, other pages are not.
2. Even so, if a page is not translated to other languages besides English, you should still get there.
3. Notice that the previous link does not have "/en-us/" as part of its link.
4. It is recommended to follow the existing codestyle and format, unless you can demonstratively and repeatedly show that the link is broken when the language directory part of the link is removed.

The bottom line (TL;DR) is that we should **always** use MS Docs links without a specific language, unless the link breaks.

illfated:
Based on my previous comment, please commit (merge) my PR code suggestion above.

illfated:
Very well, please also edit line 21:

Remarks
-------
The value of DEVPKEY_Device_ClassGuid is set by the INF ClassGUID directive that is supplied by the [**INF Version section**](https://docs.microsoft.com/windows-hardware/drivers/install/inf-addservice-directive) of the INF file that installs a device.

https://github.com/MicrosoftDocs/windows-driver-docs/pull/2041/files#diff-bcd4274f6733b04c57ac40f93bdf33dfR54-R57

`Remarks`
`-------`
`The value of DEVPKEY_Device_ClassGuid is set by the INF ClassGUID directive that is supplied by the [**INF Version section**](https://docs.microsoft.com/windows-hardware/drivers/install/inf-addservice-directive) of the INF file that installs a device.`

illfated:
Also: PR PAGE TITLE suggestion (instead of **replaced wrong link to new link #2041**): 

**Drivers/install: INF Version link correction**

Issue #2048, opened 2020-02-21
Link to dbghelp.chm


henke37:
This page is effectively a soft redirect to another resource. But said other resource is not immediately accessible. Put it online if it isn't already and add a hyperlink.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: aac181ed-4d07-c8fe-73b6-324bbb0c7dff
* Version Independent ID: 303cc2df-0bc9-7b47-16c9-c520e3c509ff
* Content: [Symbol Server API - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/symbol-server-api)
* Content Source: [windows-driver-docs-pr/debugger/symbol-server-api.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/debugger/symbol-server-api.md)
* Product: **windows-hardware**
* Technology: **windows-devices**
* GitHub Login: @DOMARS
* Microsoft Alias: **domars**

DOMARS:
@henke37 - I added links to online information for dbghelp and image help as that improves access to that information. The off line chm continues to be available. Thanks for this issue.

@RAJU2529 - Thanks for the PR, but I decided to add links to the content that is actually online. More people tend to use that instead of downloading the standalone chm file in the SDK.

PR #2051, opened 2020-02-22
added windows 10 sdk link

RAJU2529:
as per the user request #2048 .   i inserted the windows 10 sdk link 
https://developer.microsoft.com/windows/downloads/windows-10-sdk/


DOMARS:
@RAJU2529 - Thanks for the PR, but I decided to add links to the content that is actually online. More people tend to use that instead of downloading the standalone chm file in the SDK.

PR #2074, opened 2020-03-06
Multiply Assign operator missing * and formatting

dacabdi:
The multiply assign was missing the `*` and there were other minor formatting details on that area.


DOMARS:
Thanks @dacabdi, looks like the `*` was dropped in the conversion to markdown.

PR #2075, opened 2020-03-06
Fixed Open Enclave dead link

emmasab:


emmasab:
Sure, let's use this one instead!

DOMARS:
@emmasab - I went ahead and updated the link to use the new open enclave page using the text form of the URL. So will close this PR.

https://github.com/openenclave/openenclave/blob/master/docs/GettingStartedDocs/Windows_windbg.md

Thanks again for seeing this link was broken and sending in  a PR!

PR #2113, opened 2020-03-23
Updated zoom control description to caution developers on Zoom control values

microkatz:
Updated zoom control description to caution developers that the drivers may have custom zoom ranges and step values.


barrygolden:
@microkatz, thanks for the PR!

PR #2132, opened 2020-03-28
Fix a broken link

guyharris:
The URL to which NDIS_INTERFACE_INFORMATION was linked had an extra _ in it; that broke the link.


tedhudek:
Thanks for the PR @guyharris!

PR #2136, opened 2020-03-30
Update updating-device-firmware-using-windows-update.md

Cymoki:
Being a little more explicit about what it means by “device” to help avoid confusion


tedhudek:
Thanks for the PR @Cymoki!

PR #2153, opened 2020-04-07
Update creating-a-primitive-driver.md

zlockard:


zlockard:
Can we also add a link from the InfVerif 1297 error page to this new section?

tedhudek:
Thanks for the PR @zlockard!

PR #2167, opened 2020-04-12
Removing invalid characters and replacing with a colon

jasonepperly:
Removing invalid characters and replacing with a colon


EliotSeattle:
Thanks, @jasonepperly 
Appreciate the PR. We keep trying to root out the ASCII characters that creep in. We welcome the support!

PR #2176, opened 2020-04-16
Update hid-clients-supported-in-windows.md

DJm00n:
Removed leading zeroes. Usage Page and Usage ID is one byte length (unsigned 16-bit value).


DJm00n:
details on size https://docs.microsoft.com/en-us/windows-hardware/drivers/hid/hid-usages#usage-id

EliotSeattle:
Thanks, @DJm00n 
I'm closing this out. I added the spaces before the parentheses to the file, but the table is fine as is, and we don't use html tags, other than in tables. So, we'll have to do without the superscript for the asterisk.
Appreciate the PR.
Eliot

PR #2219, opened 2020-05-01
Fix some misconceptions about resources

rodwiddowson:
Fix a couple of issues in the deadlock detection documentation and note one restriction

1. You can acquire resources resursively, just not Ex when you own shared
2. There is no such thing as out of order resource release (trust me I'm a doctor :-)
3. Deadlock detection does not interact well with Kernel fuzzing.

Hope this helps


rodwiddowson:
@DOMARS thanks for reaching out.

I’m hoping that someone in the verifier team will be able to help in some way.  This isn’t a windbg issue per se..

I did mean to remove 1003 in as much as it is a meaningless bugcheck.  In hindsight this may well be an error
* either the bugcheck is still there which would be a false positive but a bug in verifier which would need to be documented none the less
* or the bug check has been removed.  I have not searched the kernel for it so I don’t know.

I think that there is a term overload here,  I took word ‘resource’ to mean ERESOURCE because in other places the documentation talks of mutex and event and so on.  But when discussing deadlocks resource can be anything which can block (including memory and threads).  

To be honest I am still researching this.  I have seen false deadlocks recently where there is not even kernel fuzzing, but in all cases I get nothing from ‘!deadlock’.

I suspect that my fear/hope was that the code has outpaced the documentation.

(With apologies for the typos, I’m typing this keyboard-impaired)

rodwiddowson:
@DOMARS I wouldn't priotitize this at this time.  It's a quirk - I spent some time looking at this verifier settings and have since come to the conclusion that this one wont work for my clients.  

Then I found the documentation and it worried me.  But others may have different expectations and experience of this verifier setting and for them this might work.

It would be interested to hear what the devs say mind.  @lorihollasch has my email address if you want to contact me

be safe

PR #2240, opened 2020-05-07
Update code-integrity-checking.md

Jreeds001:
Changed Device Guard to Hypervisor-Protected Code Integrity


EliotSeattle:
Thanks for the catch, @Jreeds001 

Updated some links while I was in there.
Eliot

PR #2249, opened 2020-05-12

AaDake:
Added new section for identifying internal PCI ports accessible to users and requiring protection


EliotSeattle:
Merging live. Thanks, @AaDake!!

PR #2254, opened 2020-05-13
Update download-the-wdk.md

Cymoki:
Updating EWDK's link to the correct page now that the target page has updated appropriately.


Cymoki:
Correcting the error with en-us being in the link path.

PR #2258, opened 2020-05-14
Cross-Certificates for Kernel Mode Code Signing: Add expiration date of cross-certificates

pmisik:
There are also open questions:

1. Do Microsoft plan to issue new cross-certificates to CA before old one expires?
2. What will happen after software vendor will sign file after given certificate expires and timestamp counter signature will contain signed time after expiration of KM certificate? It seems signtool do not verifies expiration of additional certificate provided by switch /ac.

### DIFF
--- a/windows-driver-docs-pr/install/cross-certificates-for-kernel-mode-code-signing.md
+++ b/windows-driver-docs-pr/install/cross-certificates-for-kernel-mode-code-signing.md
@@ -59,28 +59,28 @@ Follow the steps below to identify your CA, and then download the related cross-
 
 The following list contains all of the CAs that are currently supported by Microsoft for issuing SPCs for code-signing kernel-mode code.
 
-|                              CA                              |                 Root certificate thumbprint                 |                        Download link                        |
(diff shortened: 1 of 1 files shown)


tedhudek:
> 1. Do Microsoft plan to issue new cross-certificates to CA before old one expires?

No. This has been communicated to the driver community for over a year.

> 2. What will happen after software vendor will sign file after given certificate expires and timestamp counter signature will contain signed time after expiration of KM certificate? It seems signtool do not verifies expiration of additional certificate provided by switch /ac.

The timestamp proves that the certificate used for signing was valid at the time of signing, and so the signature is not considered expired. You need to use the `/kp` option with [signtool verify](https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool) for it to check the chains supplied by `/ac`. Note that given today's date, these signatures are not valid for driver signing, but are valid for /IntegrityCheck signing of user mode binaries.

tedhudek:
Thanks for the PR!

PR #2271, opened 2020-05-21
Update ksproperty-audio-volumelevel.md

darugeri:
The note section was not being rendered properly.


DOMARS:
@darugeri - Thanks for seeing this! I looked in the other audio topics and see I have a few more of these to fix. I can do that. Thanks again.

PR #2276, opened 2020-05-21
Updated Hyperlinks and Markdown Formatting

bmkaiser:
The main purpose for modifying this file was to fix a broken hyperlink to the [PnPUtil Command Syntax](https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/pnputil-command-syntax) page. By the end though, I had removed that link entirely because while this is a small article, that article was linked three times, which felt excessive.

Additionally, I found that section at the bottom on how download the utility was an HTML table. I replaced this with standard Markdown for consistency and future supportability.


tedhudek:
Thanks @bmkaiser, I think the new page looks much better!

PR #2282, opened 2020-05-26
Update irql-annotations-for-drivers.md

xiaoyinl:
The first and the third paragraphs are exactly the same. I removed the first paragraph.


EliotSeattle:
Thanks, @xiaoyinl for catching this.
We're going to leave it in because it's twice as good as other content.
Just kidding. Merged to master. Should be live shortly.
Thanks for  the PR!
Eliot

PR #2301, opened 2020-06-04
This is to clarify the use of errata 41600.

DOMARS:


DOMARS:
@tedhudek - I normally, submit via master, but as we adjust to SSO changes, it looks like I can merge into staging. Is this an acceptable workflow? If so merge away or I can.

PR #2306, opened 2020-06-05
make explicit that both steps in the driver and steps in the app are …

selerner:
…required

trying again..


EliotSeattle:
complete

PR #2307, opened 2020-06-05
Update serial-port-console-redirection-table.md

AaDake:
Updated table formatting (currently not showing correctly as a table on the public website)


DOMARS:
@AaDake - Thanks for these fixes. This table must have been missed in migration.

PR #2308, opened 2020-06-07
Fix note

NextTurn:


barrygolden:
Thanks for the PR, @NextTurn, your changes should be live soon.

PR #2311, opened 2020-06-09
Edited doc to guide correct action to perform in Configuration Steps

gahan9:
Existing action shows action to Cancel, which is not supposed to be done, hence corrected it to valid action which is to click on Apply and OK,


EliotSeattle:
Thanks for the fix, @gahan9 
Appreciate the PR.
Eliot

PR #2330, opened 2020-06-18
update language (hopefully table is not off?)

selerner:
need to check if the table is still fine...


selerner:
@EliotSeattle is there any way to see the changes when doing a "patch" change to make sure i didn't mess up the markdown table?

EliotSeattle:
@selerner 
If you go to the Files changed tab, on the top right of the form, there's an icon for raw data that looks like: <> and to the right of it there's an icon for "Display the rich diff" that looks like a page or doc. Click that for a rendering. It doesn't get everything (metadata shows weird and a couple of things like that) but it's pretty good fidelity.

selerner:
@EliotSeattle awesome! thanks for the tip!  It seems to look good, could we complete this PR when you get a chance?

EliotSeattle:
Done, @selerner

selerner:
Thank you!

PR #2343, opened 2020-06-29
Remove locale from links

NextTurn:


DOMARS:
@NextTurn - This looks good, thanks for seeing the language code!

PR #2345, opened 2020-06-30
Update windbg-what-is-new-preview.md

justanotheranonymoususer:
My WinDbg says it's of version 1.0.2001.02001. This document can be useful if kept up to date, at least to know at any given time what's the latest WinDbg version without installing it. And while at it, is there any other place which provides this information?


DOMARS:
@justanotheranonymoususer - Thanks for pointing out that are release notes were not up to date. As it turns out, version 1.0.2007.01003 has since been released and all of the features that it contains are now documented in the release notes. Thanks again for filing this PR.

PR #2360, opened 2020-07-13
Update enabling-a-kernel-mode-dump-file.md

NAO-MSFT:
Reflects a result of ICM: 65501164 to doc. It takes 3 - 4x times to generate a kernel-dump on Azure VM if DedicatedDumpFile is enabled.

### DIFF
--- a/windows-driver-docs-pr/debugger/enabling-a-kernel-mode-dump-file.md
+++ b/windows-driver-docs-pr/debugger/enabling-a-kernel-mode-dump-file.md
@@ -31,6 +31,10 @@ The settings that you select will apply to any kernel-mode dump file created by 
 
 However, these settings do not affect dump files created by the [**.dump**](-dump--create-dump-file-.md) command. See [Creating a Dump File Without a System Crash](creating-a-dump-file-without-a-system-crash.md) for details on using this command.
 
+> [!NOTE]
+> Using the DedicatedDumpFile registry entry on Azure VM may cause significant delay in creating a kernel-mode dump file.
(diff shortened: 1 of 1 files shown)


DOMARS:
This looks like useful information, but I want to see if we have this information in the docs outside of the blog post, and if we do, would prefer to use that source. I will research in the morning, so hang on for a bit.

DOMARS:
Hi @NAO-MSFT - I will follow up on email so we can discuss this. In particular, want to clarify the dump type in addition to the dump location. For example on the location, the type of storage used that is specified by , will impact the performance. Also if there is not enough space, the dump file maybe truncated, so that would shorten the time take to collect the dump.

For setting the DedicatedDumpFile value we can use this article: 

https://docs.microsoft.com/en-us/windows-server/administration/server-core/server-core-memory-dump#step-2-configure-the-destination-path-for-a-memory-dump

And on the related content, here are a few neighboring areas of docs to consider for adding what ever information we determine we want to pass on:

https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/boot-error-troubleshoot-windows

https://docs.microsoft.com/en-us/windows/client-management/troubleshoot-windows-freeze

DOMARS:
Continuing to follow up with @NAO-MSFT. It is looking like some other factor beyond changing the location of the dump file using DedicatedDumpFile is impacting the performance on the VMs.

DOMARS:
After further investigation, it is not clear that this issue is specific to the dumpfile process. As it may be a general behavior of heavy use of the paging file in certain versions of Windows, this does not appear to be specific to the kernel mode dump process. So closing this issue for now until additional information can be gathered. @NAO-MSFT, thanks for passing on what you are seeing here.

PR #2400, opened 2020-08-09
added   two relative links

RAJU2529:
as per the user report #2398, so i added two  links 
**(inf-classinstall32-section.md)**  and **(inf-addcomponent-directive.md)**


tedhudek:
Thanks @RAJU2529!

Issue #2389, opened 2020-08-01
Microsoft Doc Accessibility Workshop for Write the Docs 2020


tedhudek:
# Microsoft Doc Accessibility Workshop for Write the Docs 2020

We're so glad you could join us today!

We will be bringing the most frequently visited docs in the Windows driver documentation repo into alignment with accessibility recommendations in the Microsoft Style Guide.  By doing so, we'll set a precedent for the community, as well as Microsoft's writers.

Specifically, we're going to update occurrences of the word click, according to the following style guide recommendations:

[Describing interactions with UI](https://docs.microsoft.com/en-us/style-guide/procedures-instructions/describing-interactions-with-ui)

Here's how to proceed:

1. Look at [filenames-with-click](https://gist.github.com/tedhudek/070b08a2453314b73abc2d4434d8f6c4).
2. In the list, pick a block of files to work on.
3. Look in the discussion in this issue to ensure that no one else is already working on that block. Add a comment that you are starting work on it.
4. Either work in the browser (hint: use the `t` keyboard shortcut on GitHub to get to the file picker, where you can quickly find files) or `git clone https://github.com/MicrosoftDocs/windows-driver-docs` and work locally.
5. Open one pull request to contain all your changes. In the PR, add `#2389` in the conversation to link to this issue.
6. Add the `Write-the-Docs-2020` label to the PR so we can triage all the fixes later.

If you have questions, don't hesitate to message Ted Hudek on the Hopin platform during Writing Day.

Thank you for your help making our docs more accessible for everyone!

nitya:
I'll be starting with Block #1

nitya:
Picking up Block #2

tedhudek:
Go Nitya!!  We really appreciate your help.

nitya:
Picking up Block #3

vgiurgiu:
Picking up Block #5

nitya:
Picking up Block #4

nitya:
Picking up Block #6

nitya:
Reviewing Block #8 
Picking up Block #9

tedhudek:
Thanks to everyone for all the help!

PR #2403, opened 2020-08-09
Update using-the-netmon-tool.md

sajeanp:
updated for "click" instances; also fixed other non-standard UI interaction words


sajeanp:
#2389 
#write-the-docs-2020

PR #2416, opened 2020-08-09
Accessibility updates to replace `click` with `select`

vgiurgiu:
Fix for issue #2389 - block 5


vgiurgiu:
I'm not 100% sure about the instruction in `windows-driver-docs-pr/debugger/viewing-the-call-stack-in-visual-studio.md`. @tedhudek , it would be best if you verified this with a colleague who uses a screen reader. Thanks!

PR #2417, opened 2020-08-09
Update test-signing.md #2389 write-the-docs-2020

sajeanp:
updated for "click" instances, some minor edits too


tedhudek:
Nice going checking in Device Manager for this one to update the name of the Device status box.  Appreciate the due diligence :)

sajeanp:
After the first .md that obviously was written for an older version of Windows (which we chatted about at the time), I started testing the instructions whenever I could using the Win 10 UI :-D

PR #2425, opened 2020-08-10
Fix typo in kernel-memory-dump.md

hecongy:
Does HAL stand for hardware abstraction level instead of hardware abstraction layer? I think this might be a typo.


DOMARS:
@hecongy - I think this is a typo and it should be "abstraction layer". I can make this fix, or you could reopen this PR?

hecongy:
Thanks Don,
I closed it because I saw HAL stands for abstraction level in another doc so I think maybe I was wrong, I will reopen this one and also create a new one for that typo. Thanks for confirmation!

hecongy:
#2431 and #2432

PR #2450, opened 2020-08-24
Fixing Block 6 for #writethedocs #2389

nitya:
Fixing "click" occurences in listed files (and a few others discovered in develop/)


tedhudek:
Thanks again @nitya, these all look great!

PR #2457, opened 2020-08-25
Fix plmdebug typo.

aruproy14:
remove typo for /activateBgTask


DOMARS:
@aruproy14 - Fix looks good, thanks!

Issue #2455, opened 2020-08-24
Dead link in Using a Universal INF File section


luisalves05:
In the section "[Using a Universal INF File](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/using-a-universal-inf-file#which-inf-sections-are-invalid-in-a-universal-inf-file)" the link to "[INF Syntax](
https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-file-sections-and-directives.md)" is broken (404), I think the reason is because the docs have two section now, one for "Sections" and other for "Directives".

RAJU2529:
@PRMerger6 I am creating PR now.

RAJU2529:
@PRMerger6, verify and merge pr #2465.

EliotSeattle:
Fixed and merged

PR #2465, opened 2020-08-28
replaced invalid link with new link

RAJU2529:
as per the user report #2455 , so i replaced an invalid link to the new link.

**https://docs.microsoft.com/windows-hardware/drivers/install/general-syntax-rules-for-inf-files**


EliotSeattle:
Thanks, @RAJU2529!

PR #2472, opened 2020-09-01
append additional cause for bug check

kein284:


DOMARS:
@kein284 - The other scenario you describe looks possible. There is also this warning in KeStackAttachProcess, so I will look at adding additional causes and other wording enhancements. 

*Note   Attaching a thread to a different process can prevent asynchronous I/O operations from completing and can potentially cause deadlocks. In general, the lines of code between the call to KeStackAttachProcess and the call to KeUnstackDetachProcess should be very simple and should not call complex routines or send IRPs to other drivers.* 

In general we try to highlight the most likely causes of the bug check. 

Thanks for the PR and your interest in the stop code documentation.

DOMARS (commented):
Looks OK.

PR #2475, opened 2020-09-01
Convert DBG2 table into Markdown from Word document and update to latest version

brinez-msft:
This change submits the latest DBG2 ACPI table specification as a Markdown document (to replace the deprecated Word document version).


brinez-msft:
Note - ms.assetid field on the new DBG2 page is not filled in yet (not sure what GUID to write).

Issue #2481, opened 2020-09-04
Hyperlink issue


dcRUSTy:
[Enter feedback here]

Line Number 13 in .md file leads to hyperlink as https://windows-hardware/drivers/debugger/
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 142adea7-b7b8-4a5a-5e19-83a202ab04b2
* Version Independent ID: 489c0c22-908c-aa93-520c-836c941b5b4e
* Content: [Getting Started with WinDbg (User-Mode) - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/getting-started-with-windbg)
* Content Source: [windows-driver-docs-pr/debugger/getting-started-with-windbg.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/debugger/getting-started-with-windbg.md)
* Product: **windows-hardware**
* Technology: **windows-devices**
* GitHub Login: @DOMARS
* Microsoft Alias: **domars**

RAJU2529:
@dcRUSTy , I am fixing the hyperlink now.

dcRUSTy:
@RAJU2529 not very sure hardcoding the domain name was required or not.... removing one `/` from `//windows-hardware/drivers/debugger/` is also one of the solution

RAJU2529:
@dcRUSTy , also share the screenshot, location of **Line Number 13 in .md file leads**

EliotSeattle:
Corrected in source and merged to main branch. Should be live momentarily.

Thanks for bringing this to our attention, @dcRUSTy 
Eliot

PR #2482, opened 2020-09-04
removed invalid link, added new link

RAJU2529:
as per the user report #2481 , so I removed the invalid link and replaced the new link

new link
**https://docs.microsoft.com/windows-hardware/drivers/debugger/**


EliotSeattle:
@RAJU2529, we are using relative links to improve the offline experience. The correct link is:

 see [Debugging Tools for Windows (WinDbg, KD, CDB, NTSD)](index.md).

Because the file, debugger/index.db is in the same folder as the one we're linking from, debugger/getting-started-with-windbg.md

PR #2499, opened 2020-09-11
Update driver-isolation.md

selerner:
making it a bit more clear how you should use the api's and when to use each one


selerner:
@EliotSeattle one more ask to merge please =)

PR #2504, opened 2020-09-15
Update implementing-hardware-offloaded-apo-effects.md

kkotte:
Remove ambiguity surrounding whether the buffer size selected for low power audio is fixed at 1s.


DOMARS (approved):
@kkotte - The wording update looks good.

PR #2512, opened 2020-09-17
Fix interpretation of GetModuleFileName in sample code

jasonknichel:
Fix interpretation of GetModuleFileName in sample code


tedhudek:
Thanks for the PR, Jason, sorry for the delay processing it.

PR #2520, opened 2020-09-24
Add some debugging info for UMDF drivers

jasonknichel:
Add some debugging info for UMDF drivers


tedhudek:
Thanks for the PR Jason

tedhudek:
The update is live -- https://docs.microsoft.com/en-us/windows-hardware/drivers/install/cm-prob-failed-start#for-driver-developers

PR #2526, opened 2020-09-29
added more  abbreviations

RAJU2529:
its my own my PR , i added more abbreviations for certain commonly used technical words.


barrygolden:
Thanks for the PR, @RAJU2529. One thing of note, we want to limit this terminology table to terms used in the current topic and child topic nodes. Best practices are to expand and define terms and acronyms on first use in a topic. Occasionally, as in this case, to avoid inadvertent loss of term definitions, we will leave a terminology table in content that was ported from a white paper or other existing documentation source. 

If you have noted some terms that appear in the Firmware WEG but are missing or are incorrectly defined (such as those you have submitted previously), I'd be happy to have you submit a PR against this content.

Thanks again for your ongoing contributions to the Windows driver docs!

PR #2528, opened 2020-10-01
Update device-manager-problem-codes.md

aktsuda:
CM_FAILED_ADD is not correct. It should be CM_PROB_FAILED_ADD.


tedhudek:
Thanks Akihisa, good catch.  Those span tags don't appear on the rendered page, so I added a change to the actual text that appears on that line.

DOMARS (approved):
The fix looks good. Also checking the other codes, it looks like this is the only one in error. Thanks for the fix @aktsuda  and thanks to Ted for the rendering edits.

PR #2530, opened 2020-10-02
Updating related topics links

degranit:
The existing links are broken. The old blog was recently decommissioned, and these links are now updated to the new locations


EliotSeattle (approved):
Thanks for catching this!

PR #2534, opened 2020-10-05
Documenting power button bugcheck option

rackley096794:
Information taken from public Channel 9 episode and verified with Windows source code review.


DOMARS:
@rackley096794 - Looks good, thanks for the PR.

PR #2537, opened 2020-10-07
Update windows-kernel-mode-process-and-thread-manager.md

jasonepperly:
Updating best practices to include object manager pre and post callback routines


tedhudek:
Thanks for the PR, Jason!  The change should be on the live site within 10 min

PR #2541, opened 2020-10-07
Update how-setup-ranks-drivers--windows-vista-and-later-.md

selerner:


tedhudek:
Thanks for the PR Sebastian!

PR #2557, opened 2020-10-23
Update hid-architecture.md

DJm00n:
Clarify that you can use HidD_GetAttributes/HidD_GetManufacturerString/HidD_GetProductString for Keyboard and Mouse etc HID devices.


DJm00n:
It is actually known and used by some libs. Example: https://github.com/libusb/hidapi/blob/master/windows/hid.c#L613

billlattimer:
@DJm00n thank you for this review and update.

@tedhudek verified correct; please merge.

PR #2564, opened 2020-10-29
Critical typo fixes

skycommand:
Starting with Windows 10 v20H1, a new "Optional updates" page has been added to the Settings app, Windows Update section. However, the corresponding column in this article contained only "No", which seemed strange to me. I [found a Microsoft source that said this column must contain "N/A" and "Yes" respectively for the last two rows](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/redefining-manual-driver-updates/). (To be honest, I found that Microsoft source first.)

Also, I normalized the capitalization of the heading row of the table and eliminated verbosity. Windows 10 v20H2 is rolling out, so I reflected this fact in the last row.


barakmanor (approved):

tedhudek:
Thanks @skycommand, we're impressed that you found these nuances not long after we updated this doc page.  Appreciate the assist

PR #2569, opened 2020-11-01
Update understanding-windows-update-automatic-and-optional-rules-for-driver-distribution.md

skycommand:
- Delete two instances of the phrase "and WU has no applicable **Automatic** driver" from table row #4 (counting the heading). How can an Automatic driver be installed when Windows Update has no Automatic drivers?
- Correct all misquotations of the screenshot. Since the content of the screenshot does not appear in a GitHub diff, here is what it says:
    - "Automatic"
        - "Automatically delivered during Windows Upgrades"
        - "Automatically delivered to all applicable systems"
    - "Manual"
- In concert with the above, replace the following confusing phrases with the exact wordings of the screenshot
    - "indicated by Dynamic Update"
    - "Optional/Manual"
    - "during Windows Updates"
- Reorganize sections for better comprehension. Upon connecting a new device, the user goes (or should go anyway) through the following stages in order: Waiting for PnP, using Device Manager, using Windows Update, and finally, using the optional updates.
- Update the title to use natural capitalization and be less verbose. All of this page's siblings on Microsoft Docs use the natural case already.

### DIFF
--- a/windows-driver-docs-pr/dashboard/understanding-windows-update-automatic-and-optional-rules-for-driver-distribution.md
+++ b/windows-driver-docs-pr/dashboard/understanding-windows-update-automatic-and-optional-rules-for-driver-distribution.md
@@ -6,13 +6,13 @@ ms.topic: article
 ms.localizationpriority: medium
 ---
 
-# Understanding Windows Update Automatic and Optional Rules for Driver Distribution
+# Understanding Windows Update rules for driver distribution
 
 This article describes how you can control when Windows Update distributes your driver.
 
 When [submitting a driver to Windows Update](publish-a-driver-to-windows-update.md), the **Driver Delivery Options** section presents two radio buttons: **Automatic** and **Manual**
 
-Under the **Automatic** option there are two checkboxes: **Automatically delivered during Windows Updates** and **Automatically delivered to all applicable systems**. **Automatic** is the default setting for all new shipping labels.
+Under the **Automatic** option there are two checkboxes: **Automatically delivered during Windows Upgrades** and **Automatically delivered to all applicable systems**. **Automatic** is the default setting for all new shipping labels.
 
 ![Automatic driver promotions checkboxes](images/driver-delivery-options.png)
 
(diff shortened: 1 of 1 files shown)


tedhudek:
Hi @skycommand, thanks so much for taking the time to make these changes.  I like your new organization.  Made some small changes, including using bullets to separate 1909 and 2004 behavior. See what you think -- https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/understanding-windows-update-automatic-and-optional-rules-for-driver-distribution

skycommand:
Hello, @tedhudek. The bulleted list is fine. If anything, it helps the reader approach the subject more systematically. (At least, that's how I feel.)

But the rest of your changes are much more prominent.

### Breaking a 25-years tradition
The biggest one is the removal of this paragraph:

> In all versions of Windows 10, upon failing to find a driver, Device Manager recommends the user to try Windows Update. In version 1909 and earlier, this recommendation merely helped make sure that the Windows Update agent is in working condition. Starting with version 2004, it has become a functional recommendation for reaching the **Automatic**, and eventually, the **Manual** drivers.

I think this paragraph is important because it is talking about a serious breaking change. Perhaps I should clarify with a screenshot.
Driver update ends in Device Manager
Do you see the button that reads "Search for updated drivers on Windows Update"? This button has always been there. However, we admins had gotten used to ignoring it, because it was useless: Prior to v20H1, Device Manager would always download Manual drivers too, while Windows Update would only download it if an Automatic one was not available.

With version 20H1, we admins must break a 5-years-old habit and force ourselves to click this button because, after 25 years, Device Manager no longer performs an online search! Significant changes like this need more emphasis.

### When I say "Settings" ...

This sentence is also important:

> Starting with Windows 10, version 2004, the Windows Update agent page in the Settings app has an "Optional updates" sub-page. This page distributes **Manual** drivers.

Notice how I emphasized the fact that "Settings" is the name of an app, "Optional updates", with this exact capitalization, is a sub-page? But you've reverted it to this:

> To access **Manual** drivers, go to **Settings->Windows Update** and select **View optional updates->Driver updates**.

I had a very good reason for that change. 

1. When I tell people to "go to Settings->Windows Update", they think of a "Settings..." entry on a menu and assume "Windows Update" is a tab in a dialog box. The Fluent-style UWP is trying to break a 27-years-old GUI tradition, and it just can't be done without clarifications.
2. Your version explicitly orders them to "select View optional updates->Driver updates". Years of experience tell me that this type of wording (which employs imperative verbs) induces the belief that the "View optional updates" button is always there. I am bound to receive complaints when people see this:

[Where is the 'View optional updates' page?](https://user-images.githubusercontent.com/17097175/97924054-4c648500-1d74-11eb-994a-0f02efe77acb.png)

Oh, you think telling them to unpause the updates helps? Wait till they see this:

[Where is the 'View optional updates' page? Episode 2](https://user-images.githubusercontent.com/17097175/97925269-5b4c3700-1d76-11eb-8aa9-60f087da3830.png)

I considered writing a whole numbered list instruction set about accessing the "Optional updates" page, but eventually, thought not to overdo it.

tedhudek:
Hey @skycommand thanks for the comments.

Happy to work in the points you suggest, I just want to make them actionable for the user.

For the first one, something like: in Device Manager on v2004+, clicking the **Search for updated drivers on Windows update** button downloads any matching Manual drivers. prior to v2004, Device Manager downloads Manual drivers without clicking this button. (if I got it right, you're much more of an expert with it than I am)

For your second point (Settings), we could use what you have, but it doesn't describe the transitory nature that you explain.  Is it worth adding language to indicate that the button sometimes appears and sometimes not? (and maybe in which conditions)

skycommand:
That big button in Device Manager ("Search for updated drivers on Windows update") only opens the Settings app on the Windows Update page, nothing more. (Yes, the label is misleading.)

As for the "Optional updates" pages, I think this is what happens:

- Windows Update searches for Manual drivers during its scan, even though it doesn't install them
- Only when a Manual driver is found, the "Optional updates" page is available.

I think this page should state these facts without going into length about the necessity to un-pause Windows Update or where the link to the "Optional updates" page appears.

tedhudek:
Ok, let's try this.  For the first issue:

When it fails to find a driver, Device Manager shows a button labeled **Search for updated drivers on Windows Update**.

This button has different behavior depending on your version of Windows:

* Starting in Windows 10, version 2004, this button opens the Settings app to the Windows Update page, where the user must take action to download Manual drivers.
* In earlier versions of Windows, Device Manager downloads Manual drivers even if the user does not click this button. Clicking it does <???>

For the second issue:

Starting with Windows 10, version 2004, the Windows Update agent page in the Settings app displays **View optional updates** and then **Driver updates** only when Manual drivers are available for devices on the computer.

@barakmanor can you please verify accuracy?

skycommand:
@tedhudek For the second issue, it is all good, but I say drop the "and then Driver updates" part for brevity. After all, it is just a label. Maybe I can upload an MP4 demonstrating it. (Don't worry about the size; I'm a video encoding wizard!)

As for the first:
- Clicking "Search for updated drivers on Windows Update" in all versions of Windows 10 (and not "Starting in Windows 10, version 2004") opens the Settings app to the Windows Update page.
- Prior to v20H1, it was pointless to click it, but let's not emphasize this fact. No good comes from shaming the team in charge of Device Manager **after** they've made to pointless button useful. This isn't Wikipedia where shaming others is a virtue. (Not that I think there is the slightest chance you'd do it...😉)
- In v20H1 and v20H2, clicking it is essential, especially because, according to this article, Device Manager will not fetch anything from the Internet, no matter what.

skycommand:
And here is the video:

[Tutorial video for optional drivers.zip](https://github.com/MicrosoftDocs/windows-driver-docs/files/5479884/Tutorial.video.for.optional.drivers.zip)

tedhudek:
Thanks @skycommand, I just pushed updates that incorporate your feedback as well as feedback from my colleague @barakmanor.  Those changes should appear on the live site within a few minutes.

I appreciate your taking the time to make a video - alas, it didn't play for me.  No need to remake it however - I think we have what we need in the text and screenshots.

skycommand:
The video didn't play? 😶 Tested it in _Microsoft Edge_. Must have tested in _Movies & Video_ too.

Alright then. I had a bash. Take care.

tedhudek:
Got the video working in Edge!  Thanks for the tip.  I had to explicitly extract it, rather than opening the .zip folder and double-clicking the mp4

tedhudek:
I just added a sentence to the doc page (based on the video!) to help them find the mysterious button.  Should be live in a few minutes

PR #2571, opened 2020-11-04
Correcting typo in code window

SeanGreenbaum:
The commands being shown in the code window are in fact not C++ lines of code, but is a command to be entered into the CMD command prompt. 
Adjusting text to correctly reflect that.


tedhudek:
Thanks @SeanGreenbaum great fix

PR #2589, opened 2020-11-24
Update setting-up-a-usb-3-0-debug-cable-connection.md

josephpracharmsft:
Add links for Ice Lake and Tiger Lake info & processor model listing


DOMARS:
Waiting to hear back that this covers all that we need to.

PR #2590, opened 2020-11-24
Update gnss-driver-architecture.md

Cheoljin:
There was a customer report that a SUPL-related diagram is not clear. To reflect the feedback, I make two changes: 
(1) the sub-title with the diagram, i.e., 'Configuration', and ' Handling of user consent of NI requests". The diagram below the statement should be consistent with the subtitles. 
(2) Add SUPL CSP MSDN document. Customer wanted to have the detail while being in the context.


Cheoljin:
FYI, I am from Microsoft feature team and we have had a feedback to clarify the doc from a customer directly, so we now update this accordingly.

barrygolden (approved):

PR #2597, opened 2020-12-01
@PnpLockDown grammar: prevent(s)

loneicewolf:
prevents applications -> prevent applications (I guess?)
" entry should be set to 1 to **prevents** applications from directly "
 ->
"  entry should be set to 1 to **prevent** applications from directly  "

**1 thing now when I look closer, sorry it was maybe wrong;  I am unsure now. (It's hard to see what's correct for me anyway)** 

Please if this is incorrect, tell me that; sorry if it was


tedhudek:
Thanks William for the sharp eye :)

PR #2599, opened 2020-12-02
Update service file state APIs

jasonknichel:


tedhudek:
Thanks for the PR, Jason!   This is live now at https://docs.microsoft.com/en-us/windows-hardware/drivers/develop/driver-isolation#device-file-state

PR #2607, opened 2020-12-07
Update other-wdk-downloads.md

Cymoki:
Providing clarity


tedhudek:
Thanks for the PR @Cymoki.  This will be on the live site within the hour.

PR #2610, opened 2020-12-08
Update acpi-button-device.md

AndrewYangMSFT:
replace the term "Phones/tablets" with "Core OS". Remove HID descriptors to avoid confusion.

### DIFF
--- a/windows-driver-docs-pr/hid/acpi-button-device.md
+++ b/windows-driver-docs-pr/hid/acpi-button-device.md
@@ -20,12 +20,12 @@ ACPI device Id of a generic button device is ACPI0011. Windows loads the Microso
 
 For more information about the generic button device, visit the [Unified Extensible Firmware Interface](https://uefi.org/specifications) specifications web site, and download the *ACPI Specification Version 6.0* PDF document. Then use the left-hand pane to navigate to **Section 9.19**.
 
-## Sample buttons ACPI for phone/tablet
+## Sample ACPI button device for Windows 10 Core OS editions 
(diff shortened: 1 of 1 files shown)


tedhudek:
Assigning to Bill, the writer for hid.  My only concerns are the code identifiers on the triple-tick code fences (lines 27 and 120).  Bill, you should double-check the internal doc for valid code identifiers.  Otherwise no concerns

AndrewYangMSFT:
Thanks Ted. FYI, the sample codes on this page are ACPI Source Language (ASL), not C++.

Thanks,
Andrew

billlattimer:
@andrew, I adjusted the fences to the "console" type as there's no specific type for ACPI Source Language. Thanks for the PR!

AndrewYangMSFT:
Thanks Bill.

ASL is not new. I found this page using ASL too. There should be more.

https://docs.microsoft.com/en-us/windows-hardware/drivers/bringup/firmware-requirements-for-d3cold

Thanks,
Andrew

billlattimer:
@AndrewYangMSFT definitely not new, but also not a supported Markdown code fence type - which is why we don't see syntax color coding in the code. Having an unsupported code fence language doesn't currently generate a suggestion or warning in our publishing system, but as we continue to improve our Docs that may become a requirement - so we try to stay consistent with our guidelines. 

But I agree it's not as clear as it should be with the console type, although you call out that the fragment is ACPI in the article. I'll add the ASL type back to the code block.

PR #2612, opened 2020-12-09
Update minidriver-operations.md

DJm00n:
Fix links


tedhudek:
Thanks @DJm00n.  It's interesting, because I am currently working on a fix to update many of these links that target `drivers/ddi/index`.  I had not noticed the subsections of the `_<tech>` pages, like you used, and will incorporate that into my fix 👍

DJm00n:
Nice!

PR #2615, opened 2020-12-10
should not link to 2.3.2 version of CLI

selerner:
needs to link to the main releases page so that the user picks up the latest release.  this caused issues with another customer


selerner:
@DOMARS can you help merge this when you get a chance?  Want to avoid more partners running into this same issue

tedhudek:
Thanks Sebastian!

DOMARS:
We should probably say download the latest as that is more specific instead of appropriate.

DOMARS:
@selerner - I will work to add updated wording on the latest version and add the directory table that we discussed. :)

selerner:
Perfect thank you!!

PR #2616, opened 2020-12-10
Update static-tools-and-codeql.md

selerner:
sorry, don made a good point about using more specific language


DOMARS (approved):
Thanks, hopefully this will be helpful for the customers.

PR #2639, opened 2021-01-12
Update bug-check-0x49--page-fault-with-interrupts-off.md

AyyyAndreanna:


DOMARS:
@AyyyAndreanna - Thanks for this PR. I'm investigating the usage of this stop code in Windows.

DOMARS:
@AyyyAndreanna - Thanks for this PR. It looks like this stop code is not active at all, so I don't have a basis to describe this stop code. Your description does sound very logical, but we have had some stop codes that have been implemented in ways that don't always reflect what you would expect. Thanks for your interest in the Windows stop code documentation and good luck with your debugging work.

PR #2655, opened 2021-01-25
adjusted the table format to visible

RAJU2529:
I  verified this article in edit mode, i found one of the tables of  TdrLevel*Xxx*  is invisible to  public view, so i adjusted .


rogermachin (approved):

lorihollasch (commented):
Thanks for catching malformed table, @RAJU2529 . The table should not be a note. I'll go ahead and fix it rather than have you fuss with it. Thanks again!

PR #2657, opened 2021-01-26
Typo "Azure Blob storage"→"Azure Blob Storage"

hyoshioka0128:
https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/manage-product-submissions


tedhudek:
Thanks Hiroshi!

PR #2666, opened 2021-02-01
Update static-tools-and-codeql.md to refer to GitHub (rather than Semmle)

sj:
Hi team :wave: 

I'm the PM for CodeQL at GitHub. Really excited about the use of CodeQL in the Windows Hardware Compatibility Program! CodeQL was indeed originally developed at Semmle, which was acquired by GitHub in 2019. We've since migrated documentation and other resources to GitHub.com. I've updated the text accordingly -- I hope this is the right place to make such changes!

cc @joshbw


DOMARS:
Hi @sj ! Welcome to our docs area for drivers! We are very happy to establish communications with you, so we can stay up to date on any CodeQL doc changes. The goal will be to help our customers always find the info they need, as things move to new homes from time to time. It looks like these target link updates are live, and all ready to go?

Also, @sj, say hi to @selerner who is the primary owner of these docs. 

@selerner - do you want to look over these links, before we take them live?

selerner:
Hi there! Thanks for sending over these changes, this looks great to me.  Happy to go live.

DOMARS:
Thanks everyone! Will take live now.

PR #2678, opened 2021-02-12
fix incorrect universal inf definition

selerner:


selerner:
@tedhudek  hey ted, this PR and the other i just created selerner:patch-28 fix an incorrect defniition we have for Universal INF's.  I originally thought that "Windows Drivers" required Universal INF's, but in reality, a Universal INF is not enough to meet the "Windows Driver" reqiurements.  I'm changing some of the language to make that more clear

tedhudek:
Replaced by internal PR

PR #2679, opened 2021-02-12
typo

tjs137:


DOMARS (approved):
@tjs137 - Thanks for this typo fix! The command is indeed `C:\WINDOWS\system32> bcdedit /enum all`

PR #2681, opened 2021-02-15
Update device-poewr-management-reference.md

aktsuda:
Update the incorrect link of PoFxQueryCurrentComponentPerfState to https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-pofxquerycurrentcomponentperfstate


tedhudek:
Thanks as always Akihisa!

PR #2682, opened 2021-02-16
Breakpoint timelines are no longer automatically created

aluhrs13:
Breakpoint timelines are no longer automatically created


DOMARS (approved):
@aluhrs13 - Looks good, thanks for creating the fix in a PR!

PR #2685, opened 2021-02-17
SPCR table: Fix doc bugs

brinez-msft:
* The ‘Flow Control’ field at offset 61 has been added back. This was a documentation issue as the field is present in the source header file.
* Fix documentation error regarding the interpretation of the value for ‘Stop Bits’ field, which appears to have changed between revision 1.03 and 1.04+. (Revision 1.03 of the specification indicates: 1 = 1 Stop bit, 0,2-255 = reserved. This has been changed in revision 1.04 to: 0 = Stop bit, 1-255 = reserved.)


brinez-msft:
Ready to merge

PR #2694, opened 2021-02-20
Update native-objects-in-javascript-extensions-debugger-objects.md

yangrongwei:
add missed API.


DOMARS:
@yangrongwei - Thanks for this PR to add getModuleContainingSymbol. I'm double checking with the dev team to make sure this is OK.

DOMARS:
@yangrongwei - Looks like we are good here, so will merge in this PR. Thanks for seeing this gap in the native object docs and filing this PR.

Issue #2731, opened 2021-03-10
Checking the commit status response values


Conky5:
From step 7:
> To confirm the submission status, review the commitStatus value in the response body. This value should change from CommitReceived to CommitComplete if the request succeeds or to CommitFailed if there are errors in the request.

This seems out of sync with the status api which returns a workflow status object https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/get-product-data#workflow-status-object, and does not contain a `commitStatus` field.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0ec4d9d1-95d0-aaf4-5bc2-1fd0ff3d6ed0
* Version Independent ID: a506d854-e239-58d1-342c-99622ee5249b
* Content: [Manage product submissions - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/dashboard/manage-product-submissions)
* Content Source: [windows-driver-docs-pr/dashboard/manage-product-submissions.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/dashboard/manage-product-submissions.md)
* Product: **windows-hardware**
* Technology: **windows-devices**
* GitHub Login: @tedhudek
* Microsoft Alias: **tedhudek**

tedhudek:
Thanks for the feedback.

I think the current version is correct for this one - commitStatus for submission and workflow status are not the same thing.

Feel free to open a PR if you were thinking something different.  (Click Edit on docs.microsoft.com to start a PR.)

PR #2734, opened 2021-03-10
dashboard updates

tedhudek:


tedhudek:
Hi @conky5, this PR has proposed fixes for #2730  and #2731.  You've been through the UI recently though, it sounds like, and I never have- so is this about what you had in mind?

Conky5:
I think it would have made sense to include the headers in the `### Request header` section. The response section looks reasonable to me.

tedhudek:
Do you have a header example you'd like to add to the Request header section?  Feel free to paste here or start a new PR if so.

PR #2743, opened 2021-03-23
Update laptop-slate-mode-toggling-between-states.md

cts-davean:
Add comment indicating that CreateFile can fail with ERROR_ACCESS_DENIED when the GPIO Slate/Laptop Indicator device is already opened by another process (or the system).


tedhudek:
Thanks Dave, looks good!  Appreciate the PR

PR #2744, opened 2021-03-24
Update bug-check-0x72--assign-drive-letters-failed.md

pivotman319-new:
Add cause for ASSIGN_DRIVE_LETTERS_FAILED


DOMARS:
@pivotman319-new - Hi and thanks for this PR. To the best of my knowledge this bug check is not active in Windows, so I can't validate that this description is accurate. It would be a very reasonable description of what this bug code could be, but our goal is to describe the behavior of current day Windows 10. So will close this PR for now. If you do have additional information that should be considered let us know. Thanks for your interest in the Windows stop code documentation and creating this PR.

PR #2755, opened 2021-04-09
Update pnputil-examples.md

sldsrg:
compact layout looks better


DOMARS:
@sldsrg - Thanks for this PR. And this formatting will avoid scrolling back and forth to view the samples. I guess another option would be markdown tables, as that can enable contribution from others, but not sure how that would render, as you have less control with markdown. Thanks again for this PR, and your interest in the Windows docs!

PR #2756, opened 2021-04-11
Update bug-check-0x9f--driver-power-state-failure.md

dongpohezui:
When Parameter 1 equals 0x4, Parameter 4 should be TRIAGE_9F_PNP.


DOMARS:
@dongpohezui - Thanks for this PR! I'm doing some doublechecking, but it looks like indeed current day systems are using TRIAGE_9F_PNP for this sub type of bugcheck.

DOMARS:
@dongpohezui - Thanks for this PR, I appreciate the interest in the debugger docs and and eye for detail for the sub code values!

PR #2771, opened 2021-04-19
Updated intermediate certificate to leaf certificate in what will hap…

jgeurten:
…pen to my existing signed driver package

The timestamp must be between the ValidFrom and ValidTo dates on the leaf certificate, not the intermediate certificate. Correcting this point as a partner asked about this recently.


tedhudek:
Thanks Jordan, appreciate the PR!

PR #2786, opened 2021-05-12
Fix Built-in_drivers_for_USB_devices link

charlescao460:
Currently `Drivers for the Supported USB Device Classes` point to https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/
but I believe it means https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/supported-usb-classes#microsoft-provided-usb-device-class-drivers


tedhudek:
Thanks @charlescao460 for the sharp eye and the PR

PR #2793, opened 2021-05-16
Point link to up-to-date version of document

reflectronic:


DOMARS (approved):
Yes, this looks like the more up to date version of the Portable PDB spec. Thanks for sending this as a PR!

DOMARS:
@reflectronic - Thanks for seeing that we were linking to an older version of the portable PDB spec. I wonder if keeping the older version of the spec, serves any purpose at this point? You might consider reaching out to the .Net folks and see what there thoughts would be on removing the older copy of the spec. Thanks again for filing this PR and good luck with your .net coding.

reflectronic:
Thanks for taking a look!

The old corefx repository is mostly archived — it's used only for building & servicing older versions of .NET — so I do not think they would want to break old links and fiddle with unneeded changes to the repository.

DOMARS:
@reflectronic - That makes sense. I don't spend too much time in the .NET community, but perhaps you could consider doing a PR to the  to readme in that repo, to add text to that describes this. *"This repo is used for maintaining older versions of.NET. For the latest version see [link to latest]."*

PR #2796, opened 2021-05-19
Document fsctl-set-purge-failure-mode

rodwiddowson:
This is pretty rough and ready and includes some editorialization by me which will need to be stripped.

I think, however that it is factually accurate and should give you something to take to the developers.  It is (I hope) easier to critique something that exists rather extract the information and write it down - and this information was fresh in my head.

My apologies for the markup (also the quality of the prose - this is why I'm a dev, not a documentation person).  Also apologies for typos.


lorihollasch:
@rodwiddowson , haven't forgotten about this! The FS team is hoping to review your fine work by next week - thanks for doing the heavy lifting with the PR!

lorihollasch (approved):

PR #2827, opened 2021-06-24
Add reg key to bad memory documentation

aadriance:
This documentation is about a year out of date as Windows now stores bad pages in the registry by default.  This change adds updated details to reflect this change.


tedhudek:
Moved to internal

Issue #2832, opened 2021-06-24
Windows.Devices.Printers hyperlink links to the incorrect winrt-version


zhit-microsoft:
Currently, the link https://docs.microsoft.com/en-us/uwp/api/windows.devices.printers will redirect to
https://docs.microsoft.com/en-us/uwp/api/windows.devices.printers?view=winrt-20348, which does not show the new API interfaces.

We should change the link to the following, so that the user can see the new interfaces without additional intervention.
https://docs.microsoft.com/en-us/uwp/api/windows.devices.printers?view=winrt-insider

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d38003a1-8fba-7709-b130-57493ec9475b
* Version Independent ID: 67f50a05-3916-244c-9eb4-3a259e79bc6d
* Content: [Print support app design guide - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/devapps/print-support-app-design-guide?branch=master)
* Content Source: [windows-driver-docs-pr/devapps/print-support-app-design-guide.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/devapps/print-support-app-design-guide.md)
* Product: **windows-hardware**
* Technology: **devapps**
* GitHub Login: @barrygolden
* Microsoft Alias: **bagold**

zhit-microsoft:
The same is true for the https://docs.microsoft.com/en-us/uwp/api/windows.graphics.printing.workflow hperlink.

RAJU2529:
@zhit-microsoft . Thanks for pointing out.

RAJU2529:
@barrygolden , Sir  Please check and merge this PR #2834

barrygolden:
@RAJU2529, thanks for submitting PR #2834

We're investigating this now and will update the issue when we have a resolution.

barrygolden:
Added "?view=winrt-insider&preserve-view=true" to six print support API namespace inks in two topics:

[Print support app design guide](https://docs.microsoft.com/windows-hardware/drivers/devapps/print-support-app-design-guide#see-also)

[Print support app association](https://docs.microsoft.com/windows-hardware/drivers/devapps/print-support-app-association#see-also)

barrygolden:
@zhit-microsoft, thanks for pointing out the incorrect redirects.

[...]

PR #2834, opened 2021-06-27
removed incorrect link, added new link

RAJU2529:
as per user feedback #2832 , so I removed incorrect link and added new link.


barrygolden:
@RAJU2529, thanks for submitting PR #2834. The links have been updated with the addition of "?view=winrt-insider&preserve-view=true" to six print support API namespace inks in the following two topics:

[Print support app design guide](https://docs.microsoft.com/windows-hardware/drivers/devapps/print-support-app-design-guide#see-also)

[Print support app association](https://docs.microsoft.com/windows-hardware/drivers/devapps/print-support-app-association#see-also)

PR #2836, opened 2021-06-28
Update hid-architecture.md

DJm00n:
HID Usage Page and Usage ID are 16-bit values accodring to USB HID spec.


mhopkins-msft:
Hi @DJm00n,

Thank you for reporting this. 16-bit values are represented by 4 hexadecimal digits, so the table is correct. The USB spec is using abbreviated 8-bit values since the high-order bits are currently all zeros.

I did notice a typo in one of the values for HID UPS Battery, which I will fix.

Thanks,
 ~ *Mark*

DJm00n:
Oh, you're right.

mhopkins-msft:
@DJm00n

Thank you for taking the time to submit this, it helped me discover the typo.

Best wishes,
 ~ *Mark*

PR #2841, opened 2021-07-05
Delete **PropPageExt.dll** line from  copy files

ahidaka:
Unfortunately, Universal Windows Drivers does not allow this property page and has been removed from the sample code. Therefore, **PropPageExt.dll** cannot be copied.


DOMARS:
@ahidaka - Yes, I believe that is the case as well, and searching the sysvad sample didn't turn up anything named proppage. Thanks for seeing this error, and creating this PR with the exact solution.

PR #2846, opened 2021-07-12
Update download-the-wdk.md

0xbadfca11:
It's must be /en-us/.
Do not redirect to localized URL.
That's not exist.


tedhudek:
Hi @0xbadfca11 thanks for the suggestion. The existing URL works for me (without the lang-loc identifier).  We actually have a rule for our site to avoid lang-loc identifier and to let the user's browser make the adjustment.  Are you reporting that https://www.microsoft.com/ja-jp/software-download/windowsinsiderpreviewWDK fails?  (I see that it does.)

0xbadfca11:
Yes /ja-jp/ doesn't work.
https://aka.ms/wipwdk works well.

Issue #2845, opened 2021-07-10
link for MmGetSystemAddressForMdlSafe  is broken


rodwiddowson:
All the links for MmGetSystemAddressForMdlSafe point to `drivers/kernel/mm-bad-pointer` - and indeed the documentation for this function appears to have entirely disappeared,

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 29cf2a30-3b4d-4cec-33ff-1ed18e05c1e5
* Version Independent ID: 4c5683f8-157c-0a26-19d4-c65556e0e26b
* Content: [Using MDLs - Windows drivers](https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/using-mdls)
* Content Source: [windows-driver-docs-pr/kernel/using-mdls.md](https://github.com/MicrosoftDocs/windows-driver-docs/blob/staging/windows-driver-docs-pr/kernel/using-mdls.md)
* Product: **windows-hardware**
* Technology: **kernel**
* GitHub Login: @tedhudek
* Microsoft Alias: **tedhudek**

rodwiddowson:
OIC that page is for all the kernel macros-  you need to thumb down it to find the one you care about.

So the info is there but buried.

I leave it to you to decide whether this matters but it confused me.  And, BTW, it stops <F1 in VisualStudio> and most search engines from finding it.

tedhudek:
It's a great point Rod, thanks for reminding me.  I think we can make some improvement here.  I'll keep this issue open for a bit to keep you apprised of progress.

RAJU2529:
@tedhudek , Sir i think the below link is perfect 
**https://docs.microsoft.com/windows-hardware/drivers/kernel/mm-bad-pointer#mmgetsystemaddressformdlsafe**

RAJU2529:
I am editing this article now.

tedhudek:
Hi @rodwiddowson,

I went ahead and creating new, standalone pages for many of the macros that had been documented on the mm-bad-pointer page.  I put links to the new pages in a See Also section at the bottom of https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/mm-bad-pointer (and a Note explaining this at the top).  

Also note the new TOC in the left pane when you are on that page.

[...]

PR #2848, opened 2021-07-13
added correct link

RAJU2529:
as per user feedback #2845, so i added the below link  after  verification,
**https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/mm-bad-pointer#mmgetsystemaddressformdlsafe**
In that page, it directly pointed to **MmGetSystemAddressForMdlSafe**


tedhudek:
Thanks @RAJU2529 -- I am actually going to do a broader overhaul on these pages -- you can watch the original issue for updates.  Appreciate the PR

PR #2850, opened 2021-07-15
Fixed grammar so it reads better

d-m-grant:


DOMARS:
@d-m-grant - Thanks for the PR and I like the use of -ing gerund here. Some of our topics avoid ing as it can sound more action orientated, without it sometimes. So we have used the non ing style on purpose in some places in the docs. Thanks again for your interest in the docs and this PR!

PR #2854, opened 2021-07-17
Fix debug-universal-drivers--kernel-mode-.md

ahidaka:
Fix
The Echo driver is a simpler driver then the Sysvad audio driver.
to
The Echo driver is a simpler driver than the Sysvad audio driver.


DOMARS:
@ahidaka - Thanks for seeing this error, and sending in the fix in the form of a PR. This, then that, as they say!  : )

PR #2872, opened 2021-08-06
Update WHQL docs to refer to HLK instead of HCK

pow2clk:
The name was changed at some point from Hardware Certification Kit to Hardware Lab Kit. This updates the text to match the page it links to


tedhudek:
Thanks @pow2clk, great find.  Appreciate the PR.

PR #2875, opened 2021-08-10
Fix typo

Makonede:


tedhudek:
Sharp eye once again @Makonede, thanks!

PR #2879, opened 2021-08-20
Update windbg-restricted-mode-preview.md

rdwaite:
Fix quotation mark


DOMARS:
@rdwaite - Thanks for this PR. And we should have some sort of automation to check for these darn curly quotes.

PR #2893, opened 2021-09-09
3... not 2

jasonepperly:
There are three system defined callbacks. They are listed as SetSystemTime, PowerState, and ProcessorAdd. I suspect that one of these was added later, but the sentence before it was not updated from 2 to 3.


tedhudek:
Thanks again Jason, appreciate the sharp eye!

PR #2909, opened 2021-09-18
Improve devcon.md formatting

ratijas:
According to style guide[1], this is the right thing to do. Although, not sure about those span-id — they seem to be pretty much unused is these repositories. And the "Related topics" at the end probably needs to be reformatted as a list too.

[1]: https://docs.microsoft.com/en-us/contribute/text-formatting-guidelines


tedhudek:
Thanks @ratijas, appreciate the PR.

PR #2925, opened 2021-09-29
Update DV documentation to point to new volatile commands; Remove outdated Autofail deprecation notice

dkluemke:
Based on feedback from developers, this PR includes the following changes:
- Add links from the legacy volatile settings page to the Verifier command syntax page, which contains information about the `/now` command.
- Add a section for Windows 11 changes to the "What's New" page. Currently the only change in this section is informing about the `/now` command.
- Remove an outdated deprecation notice on the systematic fault injection page. Autofail is still actively supported.


DOMARS:
@dkluemke - I emailed over a few minor ideas for small edits. Let me know what you think.

DOMARS (approved):
Looks great!

PR #2930, opened 2021-10-03
Fix formatting in advanced-symsrv-use.md

mstange:
Not sure what happened here, it looks like the garbled result of multiple rounds of asterisk escaping. I've tried to correct and simplify it as much as possible. I kept the partial emphasis in the cache\*localsymbolcache example so that it's clear that `localsymbolcache` is a placeholder for a path and not a literal string.


DOMARS:
@mstange - Thanks for this PR! There was some automation that was run to convert the former XML to markdown. And the use of the asterisk in the server path along with it's use in Markdown can be tricky.

PR #2936, opened 2021-10-11
Add missing asterisk in Standardized Keywords' names

pcfist:
Asterisk symbol at the beginning, indicating a standard keyword, was missing in the table in `enumeration-keywords.md`. It may confuse developers not already familiar with these registry keys.

This is probably a result of automated conversion to Markdown, which caused `*` symbols to be interpreted as italic text markers.

Asterisk symbols have been re-added, and corresponding `em` tags removed.


aviviano:
Thanks for pointing out this issue @pcfist!

PR #2953, opened 2021-10-31
Remove a duplicated link from "Standard Debugging Techniques" page

masaru-iritani:
[Standard Debugging Techniques](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/standard-debugging-techniques) page has two identical links to [Crashing and Rebooting the Target Computer](crashing-and-rebooting-the-target-computer.md) page (at line 36 and 49 in standard-debugging-techniques.md). This pull request removes the second one to avoid the confusion.


DOMARS:
@masaru-iritani - As much fun as it is to crash and reboot the target PC, one link to that topic is all we need. Thanks for seeing this error, and creating a PR with exact fix! Best Regards.

PR #2977, opened 2021-11-18
debug-privilege.md: fix type

blin00:
`TOKEN_PRIVILEGE` should be `TOKEN_PRIVILEGES` (https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_privileges).

### DIFF
--- a/windows-driver-docs-pr/debugger/debug-privilege.md
+++ b/windows-driver-docs-pr/debugger/debug-privilege.md
@@ -31,7 +31,7 @@ BOOL SetPrivilege(HANDLE hToken, LPCTSTR lpszPrivilege, BOOL bEnablePrivilege)
 
     if (LookupPrivilegeValue(NULL, lpszPrivilege, &luid))
     {
-        TOKEN_PRIVILEGE tp;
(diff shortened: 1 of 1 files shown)


DOMARS:
@blin00 - It may be possible to adapt this code to use TOKEN_PRIVILEGE instead of TOKEN_PRIVILEGES, but I believe the intent is to show the use of https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-token_privileges here. 

You can see a larger example of TOKEN_PRIVILEGES here: https://github.com/microsoft/Windows-classic-samples/blob/1d363ff4bd17d8e20415b92e2ee989d615cc0d91/Samples/ManagementInfrastructure/cpp/Process/Provider/WindowsProcess.c

Thanks for looking at this code, and good luck with your debugging.

blin00:
I agree that the example looks like it's trying to use `TOKEN_PRIVILEGES`, which is why I believe the existing reference to `TOKEN_PRIVILEGE` is incorrect because `TOKEN_PRIVILEGE` doesn't exist as a type as far as I can tell :P

PR #2989, opened 2021-12-06
Update privacy-shutter-notification.md

kasumman:
Added a phrase to describe what the AVS driver should do if underlying HW does not support the privacy shutter.


barrygolden:
Thanks for the PR, @kasumman!

PR #2995, opened 2021-12-10
repair DeviceProperties_RunDLL typedef example

SamB:
It was missing a close-parenthesis, and the indentation was quite inconsistent.


tedhudek:
Thanks for the PR, Sam!  The fix should reach the live site in ~10 min

PR #2996, opened 2021-12-13
Update description regarding SD storage driver.

riwaida:
SD storage card driver is unified into single driver that names sdstor.sys


lorihollasch (approved):
Thanks very much for the PR, @riwaida!

PR #3002, opened 2021-12-21
Fix blank bullets and possible typos

alakyadav:


DOMARS:
@alakyadav - Thanks for these fixes, they look great.

PR #3027, opened 2022-01-24
Update iddcx1.4-updates-for-remote-idds.md

RogerCoote-MSFT:
Found out from UMDF team that to use UmdfHostProcessSharing you cannot have DeviceGroupId directives


lorihollasch (approved):
Thanks for the PR, @RogerCoote-MSFT!

PR #3030, opened 2022-01-27
notes about idempotency and persistence

JonathonAnderson:
We're working on debugging heap corruption that only occurs in our production IIS environment and we had to build some demo apps to intentionally cause heap corruption so we could test various scenarios related to enabling and disabling. It seemed like useful information.


DOMARS:
@JonathonAnderson - Thanks for this PR and the interest in our docs.

PR #3073, opened 2022-03-24
Remove references to Driver Package creation step.

prashantchahar:
Remove the references to Driver Package creation as that is done automatically for WDK 10 and higher.


tedhudek:
Thanks Prashant, appreciate the PR.  This should be live momentarily.

PR #3076, opened 2022-03-25
Update using-floating-point-or-mmx-in-a-wdm-driver.md

JoshCombs3:
Updating the x64 piece to include this note about ARM64 - Note: ARM64 in general is similar to AMD64 in that you don’t need to call save floating point state first. However, code that needs to be portable to x86 on kernels may still need to do that to be cross-platform.


tedhudek:
Thanks Josh, looks great.  Congrats on submitting a PR!

Fix should be live momentarily.

PR #3091, opened 2022-04-01

paphilli:
Added documentation for FailureActions directive. Minor text changes for existing text.


jasonknichel:
@tedhudek - for the "**Windows 11 Version WDK 22581**" near the end - this is a change to the OS's understanding of INFs and not an API change that would be in the WDK.  Given that, is this the right string to use?  Thanks!

tedhudek:
Should appear on the live site in ~10 minutes

PR #3092, opened 2022-04-04
Add some extra wording about the ProcessId parameter.

rodwiddowson:
The header file says 
```
        PEPROCESS ProcessId;        //  Only meaningful for FastIo locking operations.
        BOOLEAN FailImmediately;    //  Only meaningful for FastIo locking operations.
        BOOLEAN ExclusiveLock;      //  Only meaningful for FastIo locking operations.
```
From which several observations:
* ProcessId is a process _object_ not a ID.  Fixed in this PR
* ProcessId should never be set (because you cannot tell whether this is a fastio operation).  Fixed in this PR
* It is therefore impossible to send an IRP_MJ_LOCK_REQUEST on behalf of another process (for instance in a posted thread).  This is because you cannot set the ThreadID in the cbd.  Not addressed in this PR because I don't know how to.
* FailImmediately and ExclusiveLock should not be set (for the same reason as ProcessID).  Not fixed in this PR because in my experimentation you can set them so the header file might be wrong.

I sense that this PR needs more attention before its ready for prime time.

### DIFF
--- a/windows-driver-docs-pr/ifs/flt-parameters-for-irp-mj-lock-control.md
+++ b/windows-driver-docs-pr/ifs/flt-parameters-for-irp-mj-lock-control.md
@@ -50,10 +50,10 @@ Key value to be assigned to the byte-range lock.
 Starting byte offset within the file of the range to be locked.
 
 **ProcessId**  
-Opaque pointer to the process ID for the process that requested the byte-range lock.
+Opaque pointer to the process object for the process that requested the byte-range lock.  This is only valid for FastIO requests.  This field should therefore never be set when using this structure to issue a lock control function.
 
 **FailImmediately**  
-Boolean value specifying whether the lock request should fail if the lock cannot be granted immediately. This member is set to **FALSE** if the requesting thread can be put into a wait state until the request is granted or **TRUE** if it cannot.
+Boolean value specifying whether the lock request should fail if the lock cannot be granted immediately. This member is set to **FALSE** if the requesting thread can be put into a wait state until the request is granted or **TRUE** if it cannot.  
 
 **ExclusiveLock**  
 Boolean value specifying whether an exclusive lock is requested. This member is set to **TRUE** if an exclusive lock is requested or **FALSE** if a shared lock is requested.


cgallred:
Hi Rod. Thanks for looking out for our documentation. ;-)

You're correct that `ProcessId` is a pointer to a process object, not an ID.

The comments in the header about those parameters only being meaningful for FastIO operations aren't entirely correct. The internal Filter Manager routine `FltpMoveIrpToCallbackData`, where we’re transferring parameters from the IRP to the FLT_CALLBACK_DATA so we can give it to the filters, does indeed set those parameters:
```c
    case IRP_MJ_LOCK_CONTROL:

        RtlCopyMemory( &(data->Iopb->Parameters.LockControl),
                       &(irpSp->Parameters.LockControl),
                       sizeof( irpSp->Parameters.LockControl ) );
        data->Iopb->Parameters.LockControl.ProcessId = IoGetRequestorProcess( Irp );
        data->Iopb->Parameters.LockControl.FailImmediately = BooleanFlagOn( irpSp->Flags, SL_FAIL_IMMEDIATELY );
        data->Iopb->Parameters.LockControl.ExclusiveLock = BooleanFlagOn( irpSp->Flags, SL_EXCLUSIVE_LOCK );
        break;
```
Granted, that only applies to a filter _getting_ a lock control operation, not _originating_ one. You are correct that you can't send an IRP_MJ_LOCK_REQUEST on behalf of an arbitrary process. The `ProcessId` that filters below you see, and that will be passed in to the file system, will be the one that owns the thread on which you call `FltPerformSynchronousIo`/`FltPerformAsynchronousIo`.

The `FailImmediately` and `ExclusiveLock` parameters, however, can be set for a cbd issued by a minifilter and will be passed down. Here's the code that takes care of that, invoked during `FltPerformSynchronousIo`/`FltPerformAsynchronousIo` as part of preparing the cbd:
```c
    case IRP_MJ_LOCK_CONTROL:

        if (data->Iopb->Parameters.LockControl.FailImmediately) {

            SetFlag( irpSp->Flags, SL_FAIL_IMMEDIATELY );

        } else {

            ClearFlag( irpSp->Flags, SL_FAIL_IMMEDIATELY );
        }

        if (data->Iopb->Parameters.LockControl.ExclusiveLock) {

            SetFlag( irpSp->Flags, SL_EXCLUSIVE_LOCK );

        } else {

            ClearFlag( irpSp->Flags, SL_EXCLUSIVE_LOCK );
        }

        break;
```

lorihollasch:
@rodwiddowson, ditto @cgallred's thanks for your valued doc fixes/updates - always appreciated. I've approved this PR as is. Based on Christian's comments, let me know if there are further updates required in a separate issue/PR. (Thanks, Christian, for the review and info!)

cgallred:
@lorihollasch the PR added the phrase,  "This is only valid for FastIO requests.", for the `ProcessId` field, which is not correct.

lorihollasch:
@cgallred, thanks, and arg, sorry to miss that! Will remove ASAP.

rodwiddowson:
Thanks all and particularly @cgallred for the in details.

@cgallred i know this is not the correct medium (but I am unable to attend plugfest any more even virtually) so can I suggest a couple of changes?

- remove the comments for FltKernel.h
- An api to allow one to do a mj lock for a particular process (for instance in posted io) would be well cool.

thanks again to all

cgallred:
I'll file a bug to take care of the fltkernel.h comments.

Creating an API that would allow a filter to issue a lock operation for an arbitrary process would be feasible only for fast I/O. The fast I/O routines for file locks take a `PEPROCESS ProcessId` parameter and that's passed along to the lock pacakge in FsRtl, so we could put in whatever we like. But if we had to roll an IRP (either because there was no fast I/O dispatch to call or somebody denied the fast I/O), we wouldn't be able to specify the process. On the IRP path the process associated with the lock isn't a payload field of the IRP. Instead, when a file system gets `IRP_MJ_LOCK_CONTROL` it calls `FsRtlProcessFileLock` to process the specified IRP minor code. That routine gets the `ProcessId` parameter for those API calls from `IoGetRequestorProcess(Irp)`. For example, here's a snippet of `FsRtlProcessFileLock`'s processing:
```c++
    switch (IrpSp->MinorFunction) {

    case IRP_MN_LOCK:

        ByteOffset = IrpSp->Parameters.LockControl.ByteOffset;

        (VOID) FsRtlPrivateLock( FileLock,
                                 IrpSp->FileObject,
                                 &ByteOffset,
                                 IrpSp->Parameters.LockControl.Length,
                                 IoGetRequestorProcess(Irp),
                                 IrpSp->Parameters.LockControl.Key,
                                 BooleanFlagOn(IrpSp->Flags, SL_FAIL_IMMEDIATELY),
                                 BooleanFlagOn(IrpSp->Flags, SL_EXCLUSIVE_LOCK),
                                 &Iosb,
                                 Irp,
                                 Context,
                                 FALSE );

        break;
    ...
    }
```

rodwiddowson:
@cgallred got it.  I have always assumed that `IoGetRequestorProcess` gets its process from `Irp->Tail->Overlay.Thread` but I can see that might be precarious (plus adding a `PETHREAD` parameter to specify a process sucks in an API).

How about a variant of `FltAllocateCallbackData` which populates `cbd->Thread` (or even a flag for `FltAllocateCallbackDataEx`to set is from the calling thread)?  Or does that suffer from the same drawbacks (you need to know the major/minor function in order for it to make sense or that it doesn't translate 1:1 to the IRP fields?).

Anyway, thanks for your time.  I'll shut up now...

PR #3097, opened 2022-04-08
Update system-defined-device-properties2.md

DJm00n:
fix old links


barrygolden:
@DJm00n, thanks for the PR! I will add your suggested changes to an internal work item already in progress to update the links in the install docs.

DJm00n:
Thanks @barrygolden !

PR #3103, opened 2022-04-15
Fix example

GierEl180364:
Previously, the example for the INF file was incorrect, the directives were in the wrong sections.


DOMARS:
@GierEl180364 - Thanks for this PR. I checked with the dev team and although each INF file will vary depending on the specific driver, these additions are reasonable.

PR #3106, opened 2022-04-22
Update print-support-app-design-guide.md

andrew-cardy:
Updating the extension XML samples to define and use the correct namespaces, as per more recent samples and discussions with @takashi-moriyama.


andrew-cardy:
When I have some more spare time, I'm going to try and go through the document and implement the samples. I think there's one of two rough edges still, but I don't be able to confidently update the docs until I've been through the process completely at least once.

Hopefully this initial pull is fairly safe to get through - it would have helped me at least. :)

barrygolden:
@andrew-cardy, thanks for the PR!

PR #3112, opened 2022-04-25
Update accessing-device-instance-spdrp-xxx-properties.md

DJm00n:
minor nit


tedhudek:
Thanks for the sharp eye as always @DJm00n

PR #3118, opened 2022-04-29
Update devpkey-device-compatibleids.md

DJm00n:


tedhudek:
Great idea thanks @DJm00n

PR #3129, opened 2022-05-08
added windows 11

RAJU2529:
this is different PR , added windows 11


DOMARS:
@RAJU2529 - From all I know, Windows 11 continues with the use of Boot Configuration Data (BCD), so this is an appropriate update. Thanks for taking the time to do a PR for this.

PR #3137, opened 2022-05-17
Update static-tools-and-codeql.md

Nana-Akua:


DOMARS:
@Nana-Akua - Thanks for this PR. As we discussed, we need to re-test this lab, with this change to validate that everything still works end to end.

PR #3173, opened 2022-06-20
Update code-signing-cert-manage.md

isnotnick:
Reinstated links to CAs that were removed without reason: Certum, IdenTrust and Sectigo.


anaharris-ms (changes requested):
@isnotnick  These were removed because they reflected the list that is offered during registration process. Please confirm that registration should also have these links.

isnotnick:
Thanks, @anaharris-ms - yes, I believe the registration process and this page should have a complete list of all CAs that offer EV code signing, not just a small sub-set.

anaharris-ms (approved):

PR #3196, opened 2022-07-12
Update bug-check-0x6b--process1-initialization-failed.md

RubberWarrior1999:
I found this documentation on a website called Geoff Chappell, software analyst, and doing some testing I've discovered that the parameters are true, so it would be interesting to update this otherwise too bland documentation on this bugcheck. Thank you.


tedhudek:
(Closed and reopened to get a build..)

DOMARS:
@RubberWarrior1999 - Thanks for this contribution, I need to check with the dev team and confirm before we can publish this, thanks again.

PR #3210, opened 2022-07-20
remove most of use-device-guard-readiness-tool.md

rotruxal:
DG Readiness tool is deprecated. Removing most of this article. Would be nice if we could change the URL as well but I imagine that is non-trivial.


DOMARS:
Following up with Rod in email.

tedhudek:
Thanks for the PR @rotruxal - we can actually change the URL and auto-forward, but in this case it might be best to leave it the way you have it.  Maybe add a sentence that the DG Readiness tool is deprecated and that we recommend this other course of action instead.  That way people who search the web for the old tool learn what happened rather than just getting blindly forwarded.  Up to you though, just let us know...

DOMARS:
@tedhudek - I'm working with Rob on making updates to this and another page we have in test.

DOMARS:
Closing this PR as we ended up doing work in a different branch to manage this change.

PR #3217, opened 2022-08-01
Update hardware-submission-create.md to fix error in a sentence.

avaniatre4:


tedhudek:
Thanks Avani!  The update should be live in 5-10 min

PR #3220, opened 2022-08-03
Update handling-client-impersonation-in-umdf-1-x-drivers.md

terrynini:
"In UMFD versions 1.9" should be "In UMDF versions 1.9"
"UMFD" is an abbreviation for "User Mode Font Driver"


tedhudek:
Thanks Terry, appreciate the sharp eye!  The fix should appear on the live site within 5-10 minutes

PR #3236, opened 2022-08-17
Fixing /d documentation

sammallon:
The documentation for the /d options is incorrectly typed using /v


DOMARS:
@sammallon - Thanks so much for seeing this error _and_ creating a PR to fix it! Let me know if you think I should test anything else in this topic. The description for /d sounds correct, so probably just a typo in the sample text.  Thanks again!

PR #3243, opened 2022-08-22
Updated C28726

MugBergerFries:
Updated to the new format

### DIFF
--- a/windows-driver-docs-pr/devtest/28726-banned-api-usage-use-updated-function-replacement.md
+++ b/windows-driver-docs-pr/devtest/28726-banned-api-usage-use-updated-function-replacement.md
@@ -1,21 +1,26 @@
 ---
 title: C28726 warning
 description: Warning C28726 Banned API Usage.
-ms.date: 04/20/2017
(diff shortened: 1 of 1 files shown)


Rastaban (approved):

DOMARS (approved):
@MugBergerFries - This looks good, let me know when review is complete, and I can merge in.

corob-msft (approved):
Overall, LGTM. I have a suggestion that I recognize might be outside of your task scope, and also that there may be a small difference in standards between doc sets here. When we touch the C++ content, we like to bring some styling things up-to-date. One such thing is to use inline code formatting for library function names and manifest constants/preprocessor macros. In this case, that would mean wrapping CRT functions like `scanf` and `scanf_s` in ticks: \`scanf\` and \`scanf_s\`, as done in the table later in the doc. However, since this is an existing paragraph in the content that you're not otherwise touching, I can see a reasonable argument for leaving it alone.

MichaelSquires (approved):

MugBergerFries:
@DOMARS Yes, good to go, thanks!

PR #3244, opened 2022-08-22
Updated C28750

MugBergerFries:
Updated to the latest format

### DIFF
--- a/windows-driver-docs-pr/devtest/28750-banned-istrlen-usage.md
+++ b/windows-driver-docs-pr/devtest/28750-banned-istrlen-usage.md
@@ -1,28 +1,31 @@
 ---
 title: C28750 warning
 description: Warning C28750 Banned usage of lstrlen and its variants.
-ms.date: 04/20/2017
-f1_keywords: 
-  - "C28750"
+ms.date: 08/22/2022
+f1_keywords: ["C28750", "BANNED_API_USAGE_LSTRLEN", "__WARNING_BANNED_API_USAGE_LSTRLEN"]
 ---
+# Warning C28750
 
-# C28750
+> Banned usage of lstrlen and its variants: '\**function-name*' is a banned API for improved error handling purposes.
 
+This warning indicates that a function is being used that has been banned and has a more robust or secure replacement. This specific error indicates usage of lstrlen or a variant thereof.
 
-**Warning C28750: Banned API Usage lstrlen (BANNED_API_USAGE_LSTRLEN)**\
-Example output: ```Banned usage of lstrlen and its variants: *function_name* is a banned API for improved error handling purposes.```
+## Remarks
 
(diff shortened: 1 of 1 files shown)


DOMARS (approved):
@MugBergerFries - This looks good, let me know when review is complete, and I can merge in.

windows-driver-docs-pr/devtest/28750-banned-istrlen-usage.md
```diff
@@ -1,36 +1,41 @@
...
-Example output: ```Banned usage of lstrlen and its variants: *function_name* is a banned API for improved error handling purposes.```
+## Remarks
 
-This warning indicates that a function is being used that has been banned and has a more robust or secure replacement. This specific error indicates usage of lstrlen or a variant thereof. The lstrlen function and its variants are banned because they fail to transmit exceptions. This can cause error conditions to occur much later, potentially on a different thread. This makes the error conditions harder to diagnose. In addition, equivalent substitute functions can be optimized by the compiler and avoid the performance overhead of exception handlers (_try and _except blocks).  
+The lstrlen function and its variants are banned because they fail to transmit exceptions. This can cause error conditions to occur much later, potentially on a different thread. This makes the error conditions harder to diagnose. In addition, equivalent substitute functions can be optimized by the compiler and avoid the performance overhead of exception handlers (_try and _except blocks).
```
corob-msft, line 15: I'm bringing up something utterly unrelated to your scope (and to my approval) but thought I should mention it anyway. Adopting a longstanding name change might be in order: [the names `_try` and `_except` were changed to `__try` and `__except` at least 20 years ago](https://docs.microsoft.com/en-us/cpp/cpp/try-except-statement?view=msvc-170). You have to use a compiler switch that's in danger of being deprecated to get the compiler to recognize these old names.

corob-msft (approved):
LGTM. My comment is probably best addressed to the regular maintainers of this docset.

windows-driver-docs-pr/devtest/28750-banned-istrlen-usage.md
```diff
@@ -1,36 +1,41 @@
...
 ```cpp
 int example_func(char* in)
 { 
     int size = lstrlen(in);
     return size; 
 } 
 ```
+
 This is due to the use of the unsafe function lstrlen. To fix this issue, we can use the strlen as the replacement, making sure to check if the pointer is NULL: 
+
```
Rastaban, line 38: It looks like the original author forgot to put ```{}``` around the indented section of the ```if```.  Could you add them?

windows-driver-docs-pr/devtest/28750-banned-istrlen-usage.md
MugBergerFries: Interesting... I've changed them to the latter version!

windows-driver-docs-pr/devtest/28750-banned-istrlen-usage.md
MugBergerFries: Done! Good catch

Rastaban (approved):

corob-msft (approved):

DOMARS:
@MugBergerFries - Let me know if this one is ready to merge to live.

MugBergerFries:
@DOMARS Yes, good to go, thanks!

PR #3253, opened 2022-09-06
Fix grammar in Time Travel Debugging overview

bluetarpmedia:


DOMARS:
@bluetarpmedia - Thanks for this PR, and I think this wording is fine. The trace contains sufficient information to allow the code execution to be replayed, and I think this general idea comes sufficiently through in the updated wording.

PR #3259, opened 2022-09-11
Fix one incorrect word

juhovi:


DOMARS:
@juhovi - Thanks for this PR. As the registers are for use by any function, what would you think of: 

"Recall that the first four parameters are passed in registers. Since this function **uses** only four registers, none are passed on the stack."

juhovi:
I think the point here is to remind the reader that only four registers are used for passing parameters in this ABI. IMO your suggestion does not make this clearer.

DOMARS:
@juhovi _ OK, thanks. I will ponder other wording to help clarify this, but your PR is a good step in the right direction. Thanks for contributing to the docs.

PR #3261, opened 2022-09-11
Visual Studio 2022 is 64-bit

NN---:


DOMARS:
@NN--- Thanks for seeing this mistake and creating a PR to fix it. I see the Program Files folder in 2022, for Enterprise VS.

PR #3304, opened 2022-10-19
Fix typo in inline stream inspection example

maolson-msft:
The example describes how to replace a subset of p bytes in the middle of a segment of n+p+m bytes. Currently the example says you should allow the first n bytes, then block the next p bytes and inject a replacement for it, and then BLOCK the final m bytes. But clearly you'd want to ALLOW the final m bytes. I think this was just a simple typo.


aviviano:
Thanks @maolson-msft!

PR #3308, opened 2022-10-25
Clarify out-of-band policy violations for callouts

brandyking:
We have continually encountered misbehaving callout drivers in our investigations that are violating these standards. In an effort to mitigate some of these issues, we want to highlight the importance of following the points outlined.


aviviano (approved):

aviviano:
Thanks Brandy!

PR #3332, opened 2022-11-15
Update runas-system.md

baardhermansen:
Updated with markdown for Note sections. 
General small tidying up of spaces and line shifts.  Left the spanid alone, although i don't think they are relevant any more?


DOMARS:
@baardhermansen - Thanks for this PR with markdown fixes. At some point we may do a bulk removal of the span IDs. I believe they are not required, but do no harm. Thanks again!

PR #3337, opened 2022-11-18
Fix KMDF typo

wi24rd:


tedhudek:
Thanks @wi24rd for the PR!

PR #3342, opened 2022-11-27
Update flt-parameters-for-irp-mj-query-open.md

hsebs:
Fix URL.


lorihollasch (approved):
Thanks for the PR to fix this, @hsebs!

PR #3351, opened 2022-12-15
Modify the variable name of the example code in control-guid.md

bemaru:
- The NameOfTraceFlag32 variable name can better describe the range of bit 0 to bit 31. 

- Reference link https://github.com/microsoft/Windows-driver-samples/blob/45b95c2ad0f3100525deb6c0c35dcbc37e7025b5/general/PLX9x5x/sys/trace.h#L60
- It may seem trivial, but it makes the maximum number of variables 32 easier to understand.


DOMARS:
@bemaru - Thanks for this PR. As the sample code show the GUIDs starting at 1, showing 32 as the name make sense. Somewhere I think I have seen code like this that count starts at zero. In that case the 31 would of course make more sense.

PR #3352, opened 2022-12-15
cpp to console hint fix

rul3rst4:
at the docs website, it shows that the signtool is supossed to be part of c++ which is not  true. The fix is to show the appropriate "console" to the code box.


tedhudek:
Thanks Anderson, appreciate the PR.  The fix should appear on the live site within a few minutes.

PR #3359, opened 2022-12-28
Fix link in "Working with USB Pipes"

pcfist:
Link to "How to enumerate USB pipes" took to an index page, changed the address to the correct page.


tedhudek:
Thanks Roman, appreciate the fix!

PR #3362, opened 2022-12-31
Fix mistake in NotifyRouteChange2

HBSnail:
This doc is all about NotifyRouteChange2 function, so I think "NotificationHandle that NotifyUnicastIpAddressChange returns" is a mistake. It should be "NotificationHandle that NotifyRouteChange2 returns"


aviviano:
Thanks so much @HBSnail! Looks like these were some copy/paste errors. Really appreciate your help fixing these.

PR #3367, opened 2023-01-11
Update accessing-data-buffers-in-umdf-1-x-drivers.md

Shyamal-V:
Point to UMDF 2 documentation for accessing data buffers to make it easier to discover in case someone lands on this page.


Shyamal-V:
@tedhudek @wm1 fyi

tedhudek:
Thanks Shyamal, https://learn.microsoft.com/en-us/windows-hardware/drivers/wdf/accessing-data-buffers-in-umdf-1-x-drivers has the extra text now.

PR #3371, opened 2023-01-13
Moving provisioning VM note to top

Baskman-64:
Moving the note about provisioning VMs from bottom to the top of the article to avoid wasting time to developers who want to provision VMs


tedhudek:
Thanks for the PR @Baskman-64

PR #3375, opened 2023-01-19
Update Camera Shutter Privacy Guidelines 

usivagna:
Updated the shutter guidelines to reflect a relaxation of the IR shutter requirements to support scenarios like presence sensing.


barrygolden:
@usivagna, thanks for the PR.

PR #3376, opened 2023-01-19
Update-Creating-a-DVL-Outside-Of-msbuild-Section

Nana-Akua:
Updated the "Creating a DVL outside of the msbuild section" bullet point 1.


tedhudek:
Thanks for the PR @Nana-Akua!

PR #3378, opened 2023-01-20
Fixed confusing text about address spaces

ptorr-msft:
OId text was not clear, especially given previous discussion of driver pooling


Shyamal-V (approved):

tedhudek:
Thanks Peter and Shymal.  This should be live within 5 min or so

PR #3382, opened 2023-01-26
Update inf-manufacturer-section.md

zlockard:
Use consistent casing for NT architecture decorations


tedhudek:
Thanks Zac, should be live soon

PR #3389, opened 2023-01-30
Update forcing-a-system-crash-from-the-keyboard.md

ElizabethGreene:
Added note about Laptop keyboards appearing as PS/2 devices and example of alternate key mapping.

### DIFF
--- a/windows-driver-docs-pr/debugger/forcing-a-system-crash-from-the-keyboard.md
+++ b/windows-driver-docs-pr/debugger/forcing-a-system-crash-from-the-keyboard.md
@@ -42,7 +42,7 @@ If a kernel debugger is attached to the crashed machine, the machine will break 
 
 ## Defining alternate keyboard shortcuts to force a system crash from the keyboard
 
-You can configure alternate values under the following registry subkeys for different keyboard shortcut sequences to generate the memory dump file:
+You can configure alternate values under the following registry subkeys for keyboard shortcut sequences to generate the memory dump file:
 
 - For PS/2 keyboards:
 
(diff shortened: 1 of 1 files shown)


ElizabethGreene:
Apologies in advance if this is the incorrect way to submit this.  It's my first time creating a pull request to recommend a change.

DOMARS:
ElizabethGreene - PRs are a great way to propose changes to the docs. And the note about the laptop could be helpful. I was wondering if you were able to test this? 

Also, I was wondering, if the laptop if using a PS 2 keyboard driver, would you be able to set just the PS/2 keyboard registry values like this?

``` 
[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\i8042prt\crashdump]
"Dump1Keys"=dword:00000002
"Dump2Key"=dword:0000003d
```

ElizabethGreene:
Howdy. 

Yes, I've tested this, and submitted the PR after burning a silly amount of time figuring out that Laptop-PS/2 thing. :)

It is possible to set just the PS/2/i8042prt entry, just the usb/kbdhid, or both.  I've tested all three combinations.  The reason I suggested both in the PR was to cover the common real-world scenario like mine where a laptop has both keyboards.  Here's an example from my HP Zbook.  
image
My built-in keyboard is PS/2.  My wireless/main keyboard uses USB/kbdhid.  Setting both registry keys allows me to force a crash from either keyboard and saves the user confusion from the magic key combination only working on one device or the other.

HTH

ElizabethGreene:
On an unrelated note, I can't make heads-or-tails of the keyToScanTbl array.  It doesn't match any known keyboard scan code layout that I've been able to find.  Do you know how that works under the hood?  You can ping me on Teams <egreene> if that would be easier.

Thanks!

DOMARS:
@ElizabethGreene - Thanks for the information, and that makes sense. In this scenario, setting both values allows both keyboards to trigger, which is an additional useful scenario to call out.

DOMARS (approved):
@ElizabethGreene - I did a few edits to this PR. This is useful information to add, so thank you for creating this PR.

ElizabethGreene:
Thanks!  I know I'm going to forget how I made this work, and future me will be delighted to find this information.

PR #3394, opened 2023-02-02
Note for C++ function tracing preventing error

KijoKim76:
Hi 

One of my partner brought a question with error when they trace function with "::" from application written c++. In this case c++ use :: key word and made it corrupt and I found a good tip from Microsoft internal Wiki for this case and I think it would be great to be in pubic so I brought this change. I put it in PID section as there are lots of c++ inbox and 3rd party applications to debug.  

This is also requested by customer who brought this question and they want this information to be public for them ,their partners and windows eco-system as well. 

Please review it


DOMARS (approved):
@KijoKim76 - Looks good, thanks.

KijoKim76:
Great!! Thanks and I will delete  kijokim76:patch-2 branch.

PR #3409, opened 2023-02-10
Update print-support-apps.md

Mikebell-msft:


barrygolden (approved):
Updated note to align with Microsoft style guide.

PR #3415, opened 2023-02-16
Reformatting WindowsNT in bold

Baskman-64:


tedhudek:
Thanks @Baskman-64 for the sharp eye!  This change should appear on the live site momentarily.

PR #3421, opened 2023-02-28
Update bluetooth-low-energy-overview.md

DJm00n:
Fix link


mhopkins-msft:
Thanks for the fix, Dimitriy!

PR #3449, opened 2023-03-31
Add the specific Feedback Hub category to the feedback instructions

jachea:
Because the Feedback Hub category was missing, feedback was getting misdirected to my team's Feedback Hub context, Apps > All other apps


jachea:
Never mind, Developer Platform > DTrace on Windows isn't a public feedback context

PR #3450, opened 2023-04-03
Added new Provider for Code Signing Certificates

positiwise:
Our new CA Provider is registered in Mozilla CCA DB https://crt.sh/?id=7546612314


positiwise:
Adding this new Code Signing Provider.

tedhudek:
Sorry, I'm not seeing EV signing supported here.

PR #3452, opened 2023-04-04
Clarify overlapped IO on "Types of APCs" page

JasonLinMS:
Update "Types of APCs" page to clarify that overlapped IO only uses regular user-mode APCs


tedhudek (approved):

tedhudek:
Thanks Jason, looks great!  Should appear on the live site within a few minutes.

PR #3459, opened 2023-04-07
Minimal RISC-V support changes

andreiw:
The following proposed changes enable the SPCR/DBG2 to describe serial consoles on RISC-V systems that boot with ACPI (the WIP Booting and Runtime Services spec over at https://github.com/riscv-non-isa/riscv-os-a-see).

- Adds a new serial subtype to allow for RISC-V SBI (firmware) console.
- Adds a new interrupt type to allow for RISC-V interrupt controllers.
- Allow describing polling only serial ports (SBI is only polling, and of course any UART/platform may have a quirk that could break interrupt support)


andreiw:
Note: these changes do not bump the SPCR revision or modify the revision log.

DOMARS:
@barrygolden - The serial port updates look fine to me. Will assign to you for ACPI bring up.

tedhudek:
Hi @andreiw, thanks for submitting these PRs.  Unfortunately they were inadvertently merged before we had a chance to review them internally, so I reverted them internally.  I would have used the UI here to do that, but we actually had a merge conflict between the two PRs when they came in, so I didn't want to run the risk of hitting that again by reverting straight from the PRs.  

Bottom line, you may want to hang on to the source branches in your fork while we get this sorted out.  Apologies for the inconvenience, we really do appreciate your interest in updating these docs.  I will circle back internally and figure out what next steps are.

PR #3471, opened 2023-04-24
Remove "Vista"

ahidaka:
Removed "Vista" which is the source of various misunderstandings.


DOMARS (approved):
@ahidaka - Thanks for this PR, and I believe that this general information applies to many versions of Windows.

PR #3472, opened 2023-04-25
Update winacpi-osi.md

RatinCN:
`The _OSImMethod` should be `The _OSI Method`


barrygolden:
@RatinCN, thanks for catching the error and submitting a PR. The changes should be live soon.

PR #3476, opened 2023-05-01
SPCR: Forbid SGI and PPI for the GSIV field with Arm GIC

brinez-msft:
For platforms with an Arm GIC interrupt controller, the SPCR revision 4 wording does not forbid the SPCR GSIV to be set to a SGI or PPI. Neither of those Arm GIC interrupt classes is sensible for the UART. This change explicitly forbids using SGI or PPI for the UART described by the SPCR.

Contributed by: Jose Marinho <jose.marinho@arm.com>


DOMARS:
@tedhudek - Would be good to have you as a second set of eyes on approval of this PR.

PR #3488, opened 2023-05-11
Update elam-prerequisites.md

nbassett-MSFT:
VIA is not an active program.


tedhudek:
Thanks for the PR, Nick!

PR #3515, opened 2023-06-02
Update code-signing-cert-manage.md

jp-mehta:


jp-mehta:
I have added [Certera EV Code Signing Certificate](https://certerassl.com/certera-ev-code-signing-certificate) as the certificate authorities offering EV Code Signing Certificate.

anaharris-ms:
#assign @anaharris-ms

anaharris-ms:
@jp-mehta ,

Thank you for your suggestion, but we only accept certs from members of the Microsoft Trusted Root Program.

PR #3516, opened 2023-06-02
Update bcdedit--set.md

emilvissing:
typo in `bcdedit /set` for Virtual Secure Mode


tedhudek:
Thanks for the sharp eye and the PR, Emil.  The change should appear on the live site within 10 minutes or so.

PR #3551, opened 2023-06-07
Fix wrong offset for NamespaceStringLength

ArthurHeymans:


tedhudek:
Thanks Arthur, good catch!  The update should appear on the live site within a few minutes.

PR #3594, opened 2023-07-10
Fix command to examine symbols.

JustinVirtuix:
The driver is 'KmdfDriver' not 'KmdfHelloWorld'


aviviano:
Great catch, @JustinVirtuix. Thanks so much for helping us improve the docs!

PR #3601, opened 2023-07-17
Update deprecated directive documentation to show which will block signing

zlockard:


windows-driver-docs-pr/install/inf-bitreg-directive.md
```diff
@@ -15,6 +15,9 @@ ms.date: 07/08/2022
 
 # INF BitReg directive
 
+> [!CAUTION]
+> Starting with Windows 11 version 22H2, a driver package using this directive will no longer be able to receive a signature from Hardware Developer Center.
```
jasonknichel, line 19:
```suggestion
> Starting with Windows 11 version 22H2, a driver package using this directive will no longer be able to receive a [WHQL release signature](./whql-release-signature.md) from [Hardware Developer Center](./dashboard/index.yml).
```
and similar for the other caution (I'm not sure if that index.yml is correct for the link - @tedhudek ? )?

zlockard: I don't think we want to say WHQL specifically, in case there are other signature types that might be relevant. Adding the HDC link sounds good though

PR #3606, opened 2023-07-20
Update debugger-download-tools.md

erickfurukawa:
fix small typos


DOMARS (approved):
@erickfurukawa - Thanks for this PR, looks good!
I thought about saying a "set of tools" or something else (collection of utilities), but toolset should work fine here.

PR #3612, opened 2023-08-09
Improve source path docs and remove placeholder

emmasab:


DOMARS:
Thanks @emmasab - I will take a look!

DOMARS:
@emmasab - This looks great, thanks!

The only thing I see we will want to do, is to update the two topics that link to the source-code.md and change those to 
```
[Source path] (source-path.md)
```
I can merge this is in and work on those fixes.

DOMARS:
@emmasab - I will merge this in and tidy up the other links to avoid the redirects as we discussed. Thanks again for this PR!

PR #3615, opened 2023-08-14
Update directions for using AddReg to add MSISupported value

minsang-msft:
Confirmed with Plug and Play team that the intermediate AddReg steps are not necessary.


tedhudek:
Thanks Minsang, this should be live in 10 min or so

PR #3638, opened 2023-09-09
Update forcing-a-system-crash.md

PesBandi:
Fixed an off-by-one error.


DOMARS:
@PesBandi - Yes, a total of three ways to create crash dumps. Thanks for seeing this calculation totaling error, and creating a PR to fix this!

PR #3643, opened 2023-09-15
Update -dump--create-dump-file-.md

TimMisiak:
Updating the description of the "/f" flag for usermode dumps, which has not been supported for quite some time.


DOMARS:
@TimMisiak - yes, looks like that "full" user mode dumps were removed a while back. Thanks for seeing this issue, and preparing a PR to fix the docs!

PR #3662, opened 2023-09-27
Update gflags-details.md

RAJU2529:
Replaced this **HKEY_LOCAL_MACHINE** by **HKLM**


DOMARS (changes requested):
@RAJU2529 - Thanks for this PR and your careful review of our driver docs. In this case, for HKEY_LOCAL_MACHINE, I think I prefer that usage, instead of HKLM, as the longer form is what I see in regedit. Happy to add the hyphen to "program specific", so perhaps you can update this PR to just include that.

DOMARS:
@RAJU2529 - As I mentioned a while back, in this case, for HKEY_LOCAL_MACHINE, I prefer that usage, instead of HKLM, as the longer form is what I see in regedit, so will close this PR.

PR #3667, opened 2023-09-29
Add minimum OS version for all flags

emilyhobby:
Backfill minimum OS versions for all pnputil flags. Flags introduced in Win 11 21H1 or later have minimum versions documented, but pre-Win 11 flags had no specified minimum version.


tedhudek:
Thanks Emily this looks great!

PR #3675, opened 2023-10-11
Update validating-windows-drivers.md

zlockard:
Fix command line


tedhudek:
Thanks Zac, should be live soon

PR #3683, opened 2023-10-18
Update gflags.md

RAJU2529:
made Gflags to  bold, 
removed  duplicate link under see also 
**Tools included in Debugging Tools for Windows**

article link
**https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/gflags**

### DIFF
--- a/windows-driver-docs-pr/debugger/gflags.md
+++ b/windows-driver-docs-pr/debugger/gflags.md
@@ -7,19 +7,19 @@ ms.date: 03/08/2023
 
 # GFlags
 
-GFlags, the Global Flags Editor, enables and disables advanced debugging, diagnostic, and troubleshooting features. It's most often used to turn on indicators that other tools track, count, and log.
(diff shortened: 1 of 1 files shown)


RAJU2529:
@DOMARS , Below two links, are added in this article two times
**https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/gflags-examples**
**https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/global-flag-reference**

under This section includes and under see also.
I think, above links must be removed from   under see also,

DOMARS:
@RAJU2529 - Thanks for the edits, they look fine. It is OK to have links in the articles repeated in the see also. Some users use search to locate the topic, and give it a quick scan to see if it is what they want. They may skip to the bottom of the article or use the table of contents, to find something else, if their current article is not what they were after. So it would be good to restore the link in the see also in this article.

DOMARS:
@RAJU2529 - Thanks for the PR and the interest in the docs. It is OK to duplicate links in the body of the article in the see also. Some people may scan all of the way to the bottom, after reading just the first sentence or so.

PR #3687, opened 2023-10-24
Update getting-started-with-windbg.md

RAJU2529:
1. made file-path to bold
2. made windbg.exe to bold

article link **https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/getting-started-with-windbg**

after adjustment, here is the output
Screenshot 2023-10-24 161327


DOMARS:
@RAJU2529 - These edits look fine, thanks.

PR #3688, opened 2023-10-24
Update getting-started-with-windbg--kernel-mode-.md

RAJU2529:
1. made file-path to bold

article link **https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/getting-started-with-windbg--kernel-mode-**

after adjustment, here is the output
Screenshot 2023-10-24 162454


DOMARS:
@RAJU2529 - Sure, bold is fine for these paths.

PR #3690, opened 2023-10-24
Update getting-set-up-for-debugging.md

RAJU2529:
1. one paragraph split into two sentences separately
2. made few words to bold

article link
**https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/getting-set-up-for-debugging**

after adjustment,  below  is the output
Screenshot 2023-10-24 165247


DOMARS:
@RAJU2529 - These look like fine edits, thanks.

PR #3698, opened 2023-10-31
Update enabling-message-signaled-interrupts-in-the-registry.md

minsang-msft:
Clarify valid values for MessageNumberLimit registry key, and under what conditions.


tedhudek:
Thanks @minsang-msft, this will be on the live site momentarily.

PR #3700, opened 2023-11-01
Fix incorrect PnPUtil path

forderud:
PnPUtil.exe is installed to `Windows\System32` and not `Windows\System31`.

Evidence:
```
C:\>where pnputil
C:\Windows\System32\pnputil.exe
```


tedhudek:
Thanks @forderud!

PR #3702, opened 2023-11-02
Update bug-check-0x13--empty-thread-reaper-list.md

skycommand:
Added details to "bug-check-0x13--empty-thread-reaper-list.md". This page had none.

Affected Microsoft Learn articles:

- [Bug Check 0x13: EMPTY_THREAD_REAPER_LIST](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/bug-check-0x13--empty-thread-reaper-list)

### DIFF
--- a/windows-driver-docs-pr/debugger/bug-check-0x13--empty-thread-reaper-list.md
+++ b/windows-driver-docs-pr/debugger/bug-check-0x13--empty-thread-reaper-list.md
@@ -17,16 +17,21 @@ api_type:
 
 The EMPTY\_THREAD\_REAPER\_LIST bug check has a value of 0x00000013.
 
-This bug check appears very infrequently.
+This bug check **never** appears.
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


skycommand:
@DOMARS You wrote:

> Every sentence that we publish has a cost for localization, maintenance and compliance

Then let me talk about cost. Currently, there are many fake articles on the web purporting to help users "fix" the EMPTY_THREAD_REAPER_LIST bug check. If Microsoft Learn doesn't clarify its wording, it'll lose the trust of its audience. Perhaps you need an example of what happens when the trust is lost. There was once a certain app that enjoyed a 98% market share among its competitors, but in 2012 alone, its market share flatlined. Its name is Microsoft Internet Explorer. Eleven years later, [the company has pull all kinds of tricks to recover](https://www.theverge.com/23935029/microsoft-edge-forced-windows-10-google-chrome-fight), but it has been for naught.

My proposal: "This bug check is reserved for Microsoft's internal use. None of the released versions of Windows use it."

This proposal states clearly that none of the allegations of seeing this BSOD is genuine while also addresses your concern of developers re-using the bug check's code.

PR #3704, opened 2023-11-06
Update mm-bad-pointer.md

RatinCN:
Remove an unwanted character.


tedhudek:
Thanks Ratin and apologies for the delay!

PR #3706, opened 2023-11-07
Update windows-audio-architecture.md

Shailosingh:
The labels state this is the Windows 10 Audio Stack. To my knowledge, it hasn't changed for Windows 11, so I changed it to be called the Windows 11 Audio Stack.

### DIFF
--- a/windows-driver-docs-pr/audio/windows-audio-architecture.md
+++ b/windows-driver-docs-pr/audio/windows-audio-architecture.md
@@ -1,20 +1,20 @@
 ---
 title: Windows Audio Architecture
-description: This topic provided a high level summary of the Windows 10 audio architecture.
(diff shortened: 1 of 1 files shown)


DOMARS:
@Shailosingh - I too believe that the audio stack is common in Windows 10 and 11. As Windows 10 is also available, I'm thinking it would be best to say Windows 10/11 Audio Stack. Let me know if you can update the PR, or I can do that as well. Thanks for the interest in improving the Windows audio driver docs.

Shailosingh:
@DOMARS - Thank you for the feedback. I apologize for not replying sooner (I was a bit busy in real life) but, I see that you put in your suggestion yourself and I'm happy to see that. Thank you for accepting the pull request!

PR #3708, opened 2023-11-08
Update forcing-a-system-crash-from-the-keyboard.md

RAJU2529:
1. corrected grammatical errors.
2. made a few words too bold.

article link
**https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/forcing-a-system-crash-from-the-keyboard**


DOMARS (approved):
These changes look good to me, thanks. I guess both hot key and hotkey are used, but it looks like hotkey is becoming more common. Great typo catch on the keybords.

PR #3711, opened 2023-11-09
Fix incorrect grammar and unnecessary word in 0x13 bugcheck description

jkrejcha:
This adds a space between a period and the next sentence and also removes the erroneous word "appears"


DOMARS (approved):
Thanks for these fixes - looks good.

PR #3729, opened 2023-11-19
network: correct the spelling of the `NetworkAddress` keyword

jtippet:
The `NetworkAddress` keyword is irregular in that it is standardized without a leading asterisk


aviviano:
Thanks @jtippet!

PR #3731, opened 2023-11-21
Update canceling-irps.md

jeffbromberger:
Added a qualifier to make it extra clear this clause doesn't apply to all IRPs.


tedhudek:
Thanks Jeff!  This will be live momentarily.

PR #3742, opened 2023-12-02
Fix swapped MP2 and MP3 GUIDs

kasper93:
While CEA-861 defines MP2 as 0x5 and MP3 as 0x4, the GUIDs defined in ksmedia.h are in reverse order.

Update documentation to the correct GUIDs as defined in ksmedia.h.

Based on Windows SDK 10.0.22621.


DOMARS (approved):
@kasper93 - Thanks for this PR. It looks like this is an error, as the current kdmedia.h header has these in the order that your fix shows, MPEG2 and then MPEG3. 
Thanks for helping make the audio docs better!

PR #3752, opened 2023-12-19
Update mobile-broadband-device-firmware-update.md

zlockard:
Update guidance to use driver store location instead of well-known location


aviviano:
Thanks @zlockard!

PR #3754, opened 2023-12-21
Update porting-ndis-miniport-drivers-to-netadaptercx.md

amrutha-chandramohan:
Add a note mentioning iSCSI boot is not supported in Netadaptercx


aviviano:
Thanks @amrutha-chandramohan! (I just changed the alert to a text note to align with our style guide. Readers tend to miss notes when they are in alert blocks.)

PR #3756, opened 2023-12-22
Update bug-check-0x4000008a--thread-terminate-held-mutex.md

LuigiBruno:
Edited the "Cause"section.


DOMARS:
@LuigiBruno - Updates look good, thanks.

PR #3771, opened 2024-01-12
Added instruction to set the env path for dtrace

bhupeshpant19jan:
Added instruction to set the env path for dtrace. It was missing earlier and causing confusion that dtrace commands are not working.


DOMARS:
@bhupeshpant19jan - Sure the command prompt path command will work fine. Also longer term persistent setting using the environment values in settings. And I have not really looked at the new app execution alias in packaged apps, but that might work as well.
 https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-extensions

PR #3775, opened 2024-01-16
Update ndis-poll-mode.md

amrutha-chandramohan:
Add INF keyword information


aviviano:
Thanks @amrutha-chandramohan!

PR #3789, opened 2024-01-30
Add warning about "Release for Mod Write" being called as a specialkernelAPC

rodwiddowson:
Been bitten by this twice now.

This callbacks comes in as a special kernel mode APC and so can interrupt another thread even if they have called EnterCriticalRegion.  This means many common assumptions about your code stop being true.  For instance

* If a thread holds an ERESOURCE shared and the callback tries to get it Ex you will deadlock.
* "Thread local storage" should not be accessed.

I always end up only releasing locks (not hard, but bears remembering).

This PR probably requires extensive word-smithing.


lorihollasch (approved):
Thanks for the PR, @rodwiddowson! I ran it past Christian who said you are right on, and advised the changes that I plugged in. Always appreciate your willingness to share your expertise by bettering the docs for others.

PR #3795, opened 2024-02-05
Add new DirectoryNotifyFullInformation info class

cgallred:


cgallred:
@lorihollasch the links that I added are to a page that doesn't exist yet. We need to add a doc page for `FILE_NOTIFY_FULL_INFORMATION`. It wasn't clear how I could add a new page. I started by copying the page for [FILE_NOTIFY_EXTENDED_INFORMATION](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-file_notify_extended_information), but then there were a lot of extra keywords, I didn't see how it included the actual struct syntax, etc.

lorihollasch (approved):
Thanks for the PR to update this page, @cgallred. I updated it further to link to the WDK version of the structures (in ntifs.h) versus the win32 versions.

PR #3803, opened 2024-02-12
Update the docs for "ba" to include the "/w" parameter

TimMisiak:
Previously the docs only described the "/w" option in the page for bp/bu/bm, but this option applies to the "ba" style breakpoints as well (and is very useful). I copied the docs from the page for "bp" and then adapted them to be for ba instead of bp.


DOMARS:
@TimMisiak - Hi Tim, hope you are doing well. I assume you tested this type of command where a script is called, or a command is run with break on write --
`ba w 4 /w "@$scriptContents.myFunc(mymodule!globalVariable)" mymodule!globalVariable`, so happy to update the docs to include this, Thanks for the PR! 
Looks like you need to agree with the terms of docs submission for your external GitHub account. See the text in the PR on how to do that.

TimMisiak:

TimMisiak:
> where a script is called

Yup, I tested this as part of a video I recorded about conditional breakpoints: https://augmend.com/replay/4ceb728c-830e-4481-b8e9-eb1a9f041b8d#t=816

DOMARS:
Thanks again @TimMisiak, and not many PRs include a movie!

PR #3805, opened 2024-02-15
The Length parameter for IRP_MJ_QUERY_OPEN is a PULONG

rodwiddowson:
The Length parameter is a PULONG (why?) so change the words to reflect that


lorihollasch (approved):
Nice catch, @rodwiddowson - thanks for the PR!

PR #3810, opened 2024-02-20
Fix transposed links

yitzhaks:
While reading up on debugger extensions I found the TOC was broken. It took me a while to realize the links were simply switched.


DOMARS (approved):
@yitzhaks - Yes these are switched. Thanks for the eagle eye and creating a PR to te-tanlge the TOC.

PR #3811, opened 2024-02-23
Update introduction-to-the-directed-power-management-framework.md

Shyamal-V:
Clarify expectations for WDF power policy owners in the context of Dfx


tedhudek:
Thanks Shyamal for the update.  This is now live at https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/introduction-to-the-directed-power-management-framework#requirements-for-wdf-non-miniport-drivers.

PR #3817, opened 2024-03-06
Chnage 'Status Block' to 'Return Value'

rodwiddowson:
The text jarred when I read it because I thought there was an `IO_STATUS_BLOCK` involved...


lorihollasch (approved):
Thanks for catching this, @rodwiddowson! Looks like other FSCTL pages have the same issue. I've made a note to fix them within a week's time.

PR #3822, opened 2024-03-27
Update device-firmware-update-for-usb-devices-without-using-a-co-inst…

gerritsMSFT:
…aller.md

Update to be Driver Store compliant, and to Needs/Include WUDFRD service for Win11 and above

### DIFF
--- a/windows-driver-docs-pr/stream/device-firmware-update-for-usb-devices-without-using-a-co-installer.md
+++ b/windows-driver-docs-pr/stream/device-firmware-update-for-usb-devices-without-using-a-co-installer.md
@@ -154,9 +154,9 @@ ContosoFirmware.bin=1
 1 = %MediaDescription%
 
 [DestinationDirs]
-UMDriverCopy=12,UMDF ; copy to drivers\UMDF
-ContosoFirmwareCopy=12,ContosoFirmware ; copy to drivers\ContosoFirmware
-DefaultDestDir = 12
+UMDriverCopy=13,UMDF ; copy to driverstore\UMDF
+ContosoFirmwareCopy=13,ContosoFirmware ; copy to driverstore\ContosoFirmware
+DefaultDestDir = 13
 
 [UMDriverCopy]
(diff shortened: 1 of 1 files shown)


windows-driver-docs-pr/stream/device-firmware-update-for-usb-devices-without-using-a-co-installer.md
```diff
@@ -165,24 +165,37 @@ ContosoFirmwareUpdateFilterDriver.dll
 ContosoFirmware.bin
 
 [Manufacturer]
-%CONTOSO% = ContosoFirmwareUpdateFilterDriver,ntamd64
+%CONTOSO% = ContosoFirmwareUpdateFilterDriver,NT$ARCH$.6.1,NT$ARCH$.10.0...22000
```
jasonknichel, line 168: Unless you explicitly call out that this example INF is meant to be run through stampinf to resolve things like the $ARCH$, it is probably better to use a specific architecture instead of $ARCH$.

You are declaring support for 6.1 and later, but run from driverstore is not supported in general until a much later OS version than that, and UMDF drivers even later - https://learn.microsoft.com/en-us/windows-hardware/drivers/develop/run-from-driver-store

Driver packages have general support for 'run from Driver Store' starting with Windows 10 1709. However, certain device stacks may place extra restrictions on files you need to provide that plug into that stack. Some examples are these device stacks that didn't support 'run from Driver Store' until Windows 10 1803:

UMDF driver binaries: See [Restricting the Loading Location of UMDF Drivers](https://learn.microsoft.com/en-us/windows-hardware/drivers/wdf/restricting-the-loading-location-of-umdf-drivers) for more information

UEFI firmware update: See [Authoring an update driver package](https://learn.microsoft.com/en-us/windows-hardware/drivers/bringup/authoring-an-update-driver-package) for more information

jasonknichel (changes requested):

jasonknichel (approved):

PR #3824, opened 2024-03-27
Update validating-windows-drivers.md

zlockard:


tedhudek:
Looks good Zac, thanks!

PR #3830, opened 2024-04-02
Update wrmsr--write-msr-.md

gtworek:
wrmsr sets MSRs. rdmsr reads/displays them.


DOMARS (approved):
@gtworek - Thanks for seeing this error in the docs, and creating a PR to fix it.

PR #3833, opened 2024-04-09
Update dtrace.md

lzybkr:
Add single space between words.


DOMARS:
@lzybkr - Thanks for this fix! I assume this is ready to go and will push live.

PR #3837, opened 2024-05-05
Typo in javascript-debugger-scripting.md

Xacone:


DOMARS:
@Xacone - yes indeed it is WinDbg. Thanks for seeing this typo and creating a PR to fix it!

PR #3840, opened 2024-05-11
Fix a few typos (Audio section)

Blake-Madden:


DOMARS:
@Blake-Madden - Looks great, thank you! Will be nice to fix some of these embarrassing typos and mistakes in the audio topics.

DOMARS:
@tedhudek - Looks like these great fixes are now in MicrosoftDocs:staging. Let me know if there is anything else we need to do to get these fixes to main/live.

PR #3845, opened 2024-05-17
Fix a few typos (DEV section)

Blake-Madden:


DOMARS (approved):
@Blake-Madden - Looks great! It will be nice to not refer to developers as "devlopers" in our docs.

DOMARS:
Thanks again for making the docs better for the next reader!

PR #3853, opened 2024-05-30
Update bug-check-0x119---video-scheduler-internal-error.md

sirocyl:
Minor spelling change, "and" -> "an"


DOMARS:
@sirocyl - *And* it a good idea to fix this an/and typo! Thanks for helping to improve the docs.

PR #3855, opened 2024-05-31
More verbiage around MDL handling in the read/write path

rodwiddowson:
Continuation of Work Item 236672.

As per back channel conversations with @lorihollasch  and @cgallred


lorihollasch:
@rodwiddowson , thanks for these extra details. I appreciate the PR and you sharing your expertise. RE: back-channel convo, "temerity" is such a great word.

PR #3856, opened 2024-05-31
Update mbr2gpt-tool-test-guidance.md

pronichkin:
Improve readability and branding compliance by spelling out "Generation 2" explicitly instead of using acronyms where applicable.


pronichkin:
#sign-off

PR #3860, opened 2024-06-26
Update how-setup-ranks-drivers--windows-vista-and-later-.md

topone-engineer:
I found error.
0x*SS*0000000 is not correct I think
I think 0x*SS*000000 is correct.
Now I removed extra zero, please review this.

Best regards,
Brian Hernandez


tedhudek:
Thank you @topone-engineer for the sharp eye!  The fix will be live momentarily.

topone-engineer:
Thanks @tedhudek.

PR #3864, opened 2024-07-13
Update text for Bcdedit. Retire the text from NTLDR era, about boot.ini, /maxmem.

chjfth:
Retire the text from NTLDR era, about `boot.ini`, `/maxmem`.

image


chjfth:
#sign-off

DOMARS:
@chjfth - Thanks for this PR that highlights the need for updates to the memory constraints boot setting values. I ended up having to do some other updates in these topics, so I incorporated your ideas in those updates. Thanks for your interest in keeping the BCD data store value settings up to date, as it has been a while since these boot.ini settings were in wide use. Good luck in your work with Windows memory and VMWare virtual machines.

PR #3865, opened 2024-07-15
Fix a few typos

Blake-Madden:


barrygolden:
@Blake-Madden, thanks for the PR!

PR #3869, opened 2024-08-05
Remove Ethernet from Windows 10

grimbeaver:
Per the NetAdapter team at Microsoft, Ethernet is not officially supported in 2.0 for Windows 10.  The APIs required to use it were only in the "preview" version included in WDK 10.0.19041 and those APIs changed in the next release.


aviviano:
Thanks for catching this @grimbeaver!

PR #3873, opened 2024-08-27
Add warning about Not relying on SCM dependencies.

rodwiddowson:
Again this may need some word-smithing.

It turns out that you can "fltmc unload" a filter even if another service (in my case it happens to be usermode vbut I don't see why it shouldn't be another dricer) is marked as dependant upon in.

This is in contrast with "sc" and somewhat suprrised me.  So I wanted to note it.


lorihollasch (approved):
Thanks for the PR, @rodwiddowson!

PR #3875, opened 2024-08-30
Rename Bug-Check WORKER_THREAD_TEST_CONDITION to 0x163

Michael-K-Stein:
The file, and thus URL, for the BugCheck WORKER_THREAD_TEST_CONDITION had the code "0x162" instead of "0x163". The error exists only in the file name and URL, the file content itself is valid.


Michael-K-Stein:

DOMARS:
@Michael-K-Stein - Looks good and nice to have the URL properly match the topic. Thanks for the PR!

PR #3877, opened 2024-09-02
Update bug-check-0x157--kernel-thread-priority-floor-violation.md

apal-msft:


DOMARS:
@apal-msft - Looks good, thanks.

PR #3880, opened 2024-09-30
Update creating-an-inf-file-for-a-minifilter-driver.md

zlockard:
Updates for Windows 11 version 24H2


lorihollasch (approved):
Thanks for the PR with valuable updates, @zlockard!

PR #3882, opened 2024-10-14
Fix release for eSIM LPA: functionality

brunovid:
It's actually only available in 24H2 (Germanium). Not 23H2 as the documentation suggests.


brunovid:
> @brunovid, thank you for the correction! Would you like me to merge it?

Yes, please merge the change. Thank you!

PR #3885, opened 2024-10-29
Update print-support-app-association.md

riwaida:
Adding an explanation for the predefined property GUID.


barrygolden:
@riwaida, thanks for the pull request!

PR #3890, opened 2024-11-08
Adding 2025 Calendar Image

sotocama:
Adding this image to use for 2025 Calendar Release


garrettduchesne:
There is a typo reference to 2023, that should read 2024. Otherwise it looks good.

PR #3891, opened 2024-11-11
Fix minor grammar issues in dtrace.md

Cajunvoodoo:
"windows" -> "Windows"
"feedback hub" -> "Feedback Hub" 
"open source" -> "open-source" (to be consistent with other documentation) 
inconsistent imperfect tense -> present tense
Added oxford comma and a few missing periods.

### DIFF
--- a/windows-driver-docs-pr/devtest/dtrace.md
+++ b/windows-driver-docs-pr/devtest/dtrace.md
@@ -15,7 +15,7 @@ ms.date: 11/05/2024
 
 # DTrace on Windows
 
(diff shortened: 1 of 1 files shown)


Cajunvoodoo:

Cajunvoodoo:

DOMARS:
@Cajunvoodoo - Thanks for these edits! Good to have Windows, with an upper-case W, for sure. I guess when open-source is used as a noun ("I like open source") it may not be hyphenated, but I think the hyphenated version, is what we want here. Thanks again!

PR #3894, opened 2024-11-14
Update using-a-symbol-server.md

jeffstokes72:
Extra \ in the line 15 text


DOMARS:
@jeffstokes72 - Thanks, and indeed just one '\' to escape the *. Thanks for seeing this and creating a perfect PR to fix it!

PR #3895, opened 2024-11-15
Update windbg-keyboard-shortcuts-preview.md

jsschultz:
Correct several keyboard shortcuts for WinDbgX, adjust some formatting, and added new breakpoint shortcut


DOMARS:
@jsschultz  Thanks for creating a PR to updated the docs. I tried a few and these updates look good to me.

I guess it is expected behavior that if you try and change the initial break state in user mode, the keys and menu button are both just ignored.

jsschultz:
@DOMARS,

> I guess it is expected behavior that if you try and change the initial break state in user mode, the keys and menu button are both just ignored.

This is fixed now and will be available in a future release.

PR #3896, opened 2024-12-03
Update infverif_h.md to fix incorrect argument format

zlockard:
Fix incorrect argument format


DOMARS:
@tedhudek - Let me know if you would like to review this PR.

@zlockard - Minor nit would be to remove the > from the one command block - `> infverif.exe /h /v <INF file>` or add it to the others for consistency. Other then that, everything else looks good to me.

tedhudek:
Sounds good @DOMARS please go ahead with your suggestion.

PR #3900, opened 2025-01-06
Describe the /f parameter of .restart

WelliSolutions:
.restart has a /f parameter which restarts a process that was attached to. This was not documented.


DOMARS:
@WelliSolutions - First off thanks for creating this PR to expand the docs and include the /f force option. I checked with the dev team, and as this option may not work properly in all of the debugging environments, such as KD, CDB, classic debugger, and its behavior can be impacted by other factors, at least for now, we will not be adding it to the docs. There are certainly things that the debugger can do, and that can be done with the debugger, that are more complex and more advanced. But in general, we focus the docs on basic commands that we expect to work across most use case situations. Thanks again for creating the PR, and highlighting this possible docs addition.

PR #3901, opened 2025-01-06
Add new per-device DMA remapping opt-in mechanism information

serendipitee:
This updates the enabling DMA remapping page with new information on the latest per-device registry opt-in mechanism. Additionally corrected some outdated information on the old per-driver opt-in method.


barrygolden (approved):
Update titles and headings for consistency

PR #3903, opened 2025-01-09
Rectify malformed `br` elements

Rageking8:
Fix issue where `br` elements are malformed due to some typo.


tedhudek:
Thanks for the PR!

PR #3905, opened 2025-01-24
Update download-the-wdk.md

RatinCN:
SDK 10.0.26100.2454 is coming.


tedhudek:
Thank you for the sharp eye @RatinCN!

PR #3916, opened 2025-02-21
Fix typo

changeworld:
VMWare -> VMware


DOMARS (approved):
@changeworld - Thank for seeing this error and creating a great PR to fix it!

PR #3925, opened 2025-03-07
PartialCheckedBuild: Remove dead link

nir9:
Additionally, it seems that the checked kernel and HAL are no longer provided, I couldn't find them in the WDK installation directory, and the documentation about this in the Microsoft docs seems to be lacking. If so, another approach is to completely delete this page (I can amend this if necessary).


DOMARS:
@nir9 - Thanks for this PR. And yes, I would agree that it is probably time to move this topic and a few others to the previous version of Windows area. I will work on that.

PR #3931, opened 2025-03-26
Update code-signing-reqs.md, remove Entrust

scheibling:
Replace the link to Entrust with Sectigo, since Entrust is no longer issuing certificates and refers to Sectigo. See https://store.entrust.com/code-signing-solutions

A point of clarification here though, is this list exhaustive? It is not made clear if these are the *only* valid certificate authorities for EV Certificates, or if you can use any authority in the MS Trusted Root Program, a bit of clarification here would help. Another section of the documentation links to the Trusted Root Program list.

### DIFF
--- a/windows-driver-docs-pr/dashboard/code-signing-reqs.md
+++ b/windows-driver-docs-pr/dashboard/code-signing-reqs.md
@@ -21,7 +21,7 @@ For more extensive information on driver signing requirements see the following 
 Code signing certificates can be purchased from one of the following certificate authorities:
 
 - [DigiCert code signing certificate](https://order.digicert.com/step1/code_signing)
-- [Entrust code signing certificate](https://www.entrust.com/products/digital-signing/code-signing-certificates)
+- [Sectigo code signing certificate](https://www.sectigo.com/ssl-certificates-tls/code-signing)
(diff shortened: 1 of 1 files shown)


scheibling:
> This same link to Entrust also exists in this article:
> 
> https://learn.microsoft.com/windows-hardware/drivers/dashboard/code-signing-cert-manage

This list answers my follow up question I guess, would it make sense to sync them or add some clarification that the CAs in the original section aren't the only valid ones? The phrasing currently suggests that it's an exhaustive list

scheibling:
I've moved the list from "Manage code signing certificates" to Requirements and added a reference consistent with the other usages around the docs, so there should only be one list left to maintain. Also added the EV prefix there to clarify

I'd like to add the below phrasing to the description on code-signing-reqs#where-to-get-ev-code-signing-certificates as well, but I'm not quite sure if that is the case or if there are exclusions for various reasons so I'm skipping that for now.

"or any of the trusted vendors in the [Microsoft Trusted Root Program](/security/trusted-root/participants-list)"

scheibling:
Thanks Mark! I'm quite happy with it on my end, so you can go ahead. 

Have a nice weekend!

PR #3935, opened 2025-04-15
Fix typo: VSCode -> VS Code

changeworld:
The term `VSCode` is not the official abbreviation. Per the [Visual Studio Code documentation](https://code.visualstudio.com/docs/setup/setup-overview), the proper form is `VS Code`.


DOMARS:
@changeworld - Thank you for seeing this mistake and creating a great PR to fix it!

PR #3938, opened 2025-04-30
WinDbg: Fix getting started installation legacy

nir9:
Problem: The getting started guide has some legacy from when the standard way to obtain WinDbg was via the debugging tools.

Solution: Modernize it to refer to the standalone WinDbg


DOMARS:
@nir9 - Thanks, good catch! No longer any need to find the location of the .exe with the modern debugger.

PR #3939, opened 2025-05-02
WinDbg: Fix legacy on ENTER command documentation

nir9:
Problem: The ENTER command remarks refer to an option which disables repeating the last command on ENTER, this option indeed exists on the legacy WinDbg, but not on the modern version. The modern one seems to just have this enabled by default, with no option to disable it in the UI.

Solution: Remove the sentence regarding this.

Also, I wasn't sure if by convention documentation referring to legacy/classic WinDbg should remain with a side note that it isn't relevant for the new version or just removed completely from the online docs. I can amend this in case this does need to remain.

### DIFF
--- a/windows-driver-docs-pr/debuggercmds/enter--repeat-last-command-.md
+++ b/windows-driver-docs-pr/debuggercmds/enter--repeat-last-command-.md
@@ -36,9 +36,7 @@ ENTER
 
 ## Remarks
 
-In CDB and KD, pressing the ENTER key by itself at a command prompt reissues the command that you previously entered.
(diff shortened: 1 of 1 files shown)


DOMARS (approved):
@nir9 - Thanks for this PR! The goal of the command docs, is to avoid UI references as much as possible and leave that to the current WinDbg UI described in "debuggercmds", or in the "Classic" WinDbg docs over in "debugger".  
In addition to the UI, I checked .settings and workspaces, and this behavior does not seem settable in the present day WinDbg, so great to reflect that in this topic.

nir9:
@DOMARS Noted, thanks

PR #3940, opened 2025-05-05
WinDbg: Fix some incorrections in the unassemble doc

nir9:
Problem: The ```u``` command documentation contains some incorrect claims regarding the behavior of running the command with no parameters and using the ```ub``` variant.

Solution: Fix the incorrections.

### DIFF
--- a/windows-driver-docs-pr/debuggercmds/u--unassemble-.md
+++ b/windows-driver-docs-pr/debuggercmds/u--unassemble-.md
@@ -15,7 +15,7 @@ api_type:
 # u, ub, uu (Unassemble)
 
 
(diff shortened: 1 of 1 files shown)


DOMARS (approved):
Thanks @nir9 for this PR! The nine bytes was a reference to Itanium, which is certainly not a common use case scenario. Also great to fix the mistake in markdown for `u`.

nir9:
@DOMARS Interesting, BTW regarding [daec5d5](https://github.com/MicrosoftDocs/windows-driver-docs/pull/3940/commits/daec5d5d015c42e9595f86079117e04d07dee717), note that it also works with the same default of 8 instructions on ARM64.

PR #3941, opened 2025-05-21
Update bug-check-0xb8--attempted-switch-from-dpc.md

apal-msft:


DOMARS (approved):
Yes a *deferred* DPC call. Thanks for seeing this error and creating a PR to fix it!

PR #3955, opened 2025-06-17
.send_files is only supported in live user mode sessions.

josephpracharmsft:
.send_files is only supported in live user mode sessions.


DOMARS:
We have a new publishing process, so this may take a day to move from main to live. Thanks again for the PR!

PR #3956, opened 2025-06-19
Update ntintsafe-design-guide.md

BrianBu01:
Removing language around "Integer Underflow".  This code does not have any logic to handle "Integer Underflow".

If this code does handle "Integer Underflow", I would love to see an example.

### DIFF
--- a/windows-driver-docs-pr/kernel/ntintsafe-design-guide.md
+++ b/windows-driver-docs-pr/kernel/ntintsafe-design-guide.md
@@ -1,15 +1,15 @@
 ---
 title: Using Safe Integer Functions
-description: The ntintsafe library provides a set of C functions that perform safe integer arithmetic operations with bounds checking to prevent overflows and underflows in kernel-mode code.
+description: The ntintsafe library provides a set of C functions that perform safe integer arithmetic operations with bounds checking to prevent overflows in kernel-mode code.
 ms.date: 09/28/2023
 ms.topic: concept-article
 ---
 
 # Using Safe Integer Functions
 
-One way to minimize security problems is to prevent integer overflows and underflows. Integer overflows occur when the result of an arithmetic operation is larger than the memory space of the data type that is set to receive it. This results in the truncation of the integer and an incorrect result. An underflow occurs when an operation, usually subtraction, gives an incorrect result. Casting between two data types can also cause incorrect results due to truncation of a result that does not fit the new memory space.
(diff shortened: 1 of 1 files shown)


v-dirichards:
@lorihollasch

Can you review the proposed changes?
 
**Important**: When the changes are ready for publication, adding a `#sign-off` comment is the best way to signal that the PR is ready for the review team to merge.
 
#assign: @lorihollasch
#label:"aq-pr-triaged"
@MicrosoftDocs/public-repo-pr-review-team

lorihollasch:
Thanks for your PR and interest in this page, @BrianBu01. The Safe Integer functions do indeed flag "underflow" issues; for example, `NTSTATUS status = RtlIntSub(INT_MIN, 1, &result);` should return a result = -2,147,483,649 but this number can't be represented in a 32-bit field. Instead, `result` is undefined and `status` is STATUS_INTEGER_OVERFLOW, which is the status value returned to report both overflows and underflows.

That said, "underflow" is inadequately described. I'm going to close this PR, but have filed an internal work item to update the page. I'll also add examples to show overflow and underflow.

BrianBu01:
@lorihollasch, I think your example is not an underflow, but is an overflow.

Int32 Min Value is -2147483648 in decimal and 10000000000000000000000000000000b in two's compliment binary.   The most significant bit represents the sign, in this case being negative.

`
   /* Decimal UInt32 */

         2147483648
    -                      1
    _________________
        2147483647

   /* Unsigned Integers */

        10000000000000000000000000000000b  
   -    00000000000000000000000000000001b
   _________________________________________________
         01111111111111111111111111111111b

   /* Decimal Overflow on Int32 */

        -2147483648
    -                      1
    _________________
        2147483647

   /* 
    * 2's Compliment Binary - Signed Integers 
    * I'll represent this binary [sign] [number]
    */

        [1] [0000000000000000000000000000000] b  
   -   [0] [0000000000000000000000000000001] b
   __________________________________________________
        [0] [1111111111111111111111111111111] b
`

When you minus 1 from unsigned min value, there is a borrow all the way from the most significant bit.  This borrow ends up occurring on the sign bit.  This means the arithmetic operation on the number erroneously effected the sign.  When a data structure's boundary is breached it is an overflow.  IE: The arithmetic operation overflowed the number boundary and altered the sign boundary. 

https://en.wikipedia.org/wiki/Two%27s_complement#Subtraction

I am now also confused about why RtlUIntSub can return a STATUS_INTEGER_OVERFLOW.  ISO/IEC 14882:2024 says in regards to integer logic, "Unsigned arithmetic does not overflow.  Overflow for signed arithmetic yields undefined behavior". 

Since I am not seeing the underflow on a signed integer, can you please define it?

PR #3982, opened 2025-08-25
Clarify Direct3D 11 Shared Primaries

sooknarine:
This doc mentions shared resources and primaries, it's unclear whether shared primaries should be treated more like shared resources or more like primaries.

Clarify that shared primaries follow the rules of shared resources.  Only non-shared primaries have exceptions.

- "This restriction applies even to shared resources or to" -> "This restriction applies even to shared resources, shared primaries, or to"
- "Primaries have the following notable exceptions" -> "Shared primaries follow the rules for all shared resources. Non-shared primaries have the following notable exceptions"

Based on email thread "RE: CreateResource(D3D11) for primaries" with @jenatali


jenatali (approved):

lorihollasch:
#sign-off

PR #4017, opened 2026-02-04
Updated text to indicate tool will be coming soon

elliesekineMS:


mhopkins-msft (changes requested):
We don't document things that are not released or are coming soon. These are promises and you never know how things will turn out. Only document the platform as it exists. When new things are released, you can update the docs then.

PR #4022, opened 2026-03-06
Fix: bug in creating-an-interrupt-object.md

frankh0:
The documentation was wrong here. It is not allowed to call WdfDeleteObject for a WDF_INTERRUPT object. This is documented under WdfDeleteObject. It also produces a crash if the WDF verifier is enabled and you call WdfDeleteObject for a WDF_INTERRUPT object.


ttorble:
@tedhudek 

Can you review the proposed changes?  

IMPORTANT: When the changes are ready for publication, adding a `#sign-off` comment is the best way to signal that the PR is ready for the review team to merge.  

#label:"aq-pr-triaged" 
@MicrosoftDocs/public-repo-pr-review-team

PR #4027, opened 2026-03-28
Fixed a minor typo for one of the links

pedasmith:
The link for "For policy requirement"... incorrectly had a CR between the square-bracket and parentheis parts of the link. 

This causes the link to be displayed as text and not as a real link


ttorble:
Thanks @pedasmith

PR #4051, opened 2026-08-15
Update 'Analyzing Mobile Broadband' instructions for modern wireshark

gpotter2:
Wireshark `etwdump` has been updated quite a bunch recently, and the instructions to collect MBIM infos should be updated to reflect that. (https://learn.microsoft.com/en-us/windows-hardware/drivers/network/analyzing-mobile-broadband-logs-in-wireshark)

The modern interface looks like this:
wireshark-mbb-logs0 5 PNG
wireshark-mbb-logs1


gpotter2:
Hi @aviviano and sorry for the ping ! Thought you might be interested by this since you wrote the original article :)

PR #4052, opened 2026-08-16
btvs: document keys to revert the configuration

gpotter2:
Clicking on the `btvs.exe` user interface sets some registry keys which may not be reverted (if closed unexpectedly for instance). It sounds very useful to document those so that a revert is possible.


gpotter2:
Hi @mhopkins-msft and sorry for the ping. I thought you might be interested in this because you're one of the last to have updated this file. I also wanted to warn you guys that wireshark has removed the `TCP@` in 4.7.3+ (https://gitlab.com/wireshark/wireshark/-/work_items/21232) so note that `btvs.exe` will break in a near future.

If it's of any help, wireshark's `etwdump`, now supports binding to the BTH-BTHPORT provider (https://gitlab.com/wireshark/wireshark/-/merge_requests/26193) so it could be an alternative, but it likely would benefit having some Microsoft knowledge.

Have a good week

PR #4053, opened 2026-08-21
Add ComputeAccelerator device setup class

rhythm1705:
## Summary
- add the ComputeAccelerator setup class for neural processors
- document its use for NPU and other MCDM-based drivers
- clarify that some NPUs can be integrated with GPUs under the Display class

## Tracking
- ADO work item 63697395


brdegeer:
#sign-off

